PR #8, opened 2015-04-24
Search sessions with year, id, track, title, focus

DanielKao:
Users can search sessions using query string like scroll,iOS,2013

### DIFF
--- a/WWDC/Session.swift
+++ b/WWDC/Session.swift
@@ -23,6 +23,12 @@ struct Session {
     var year: Int
     var hd_url: String?
     
+    var sessionDescription: String {
+        get {
+            return year.description + "," + id.description + "," + track + "," + title + "," + ",".join(focus)
+        }
+    }
+    
     var progress: Double {
         get {
             return DataStore.SharedStore.fetchSessionProgress(self)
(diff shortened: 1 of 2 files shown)


insidegui:
I will not merge this since It conflicts with the existing qualifier-based search. You can already search by year, track, title and focus like this: "year:2013 focus:ios scroll". Thanks anyway.
captura de tela 2015-04-25 as 00 10 37

DanielKao:
I thought using a simple query string is more intuitive than the qualifier-based search.
Thanks anyway.

gprasant:
@insidegui 

 Does the qualifier based search exist in the latest version available for download ? 

I'm using v4.3.1 and I don't see the qualifier based search working.

insidegui:
@gprasant It doesn't, it got removed in favor of the filtering UI.

gprasant:
Thanks for the quick reply. I think the qualifier based search still has value.. Especially for searching by title. I had a hard time finding the notifications talk from WWDC 2016. (I didn't know what its title was). 
And looking for "Notifications"  gave me so many results that it wasn't useful.

insidegui:
I think this will be solved with [Version 5](https://github.com/insidegui/WWDC/tree/version5), I'm planning on having an option to search by title and I want to use SearchKit for ranking (not sure whether it'll be ready before this year's WWDC tho)

gprasant:
Thanks.

PR #19, opened 2015-04-28
Improve session default sorting. Sort by year, then by session number.

jakemarsh:
While using the app, I found that it simply takes Apple's sorting of the sessions (which is seemingly random) and passes it right along.

IMHO sorting by year in descending order, then sorting by session in ascending order is the most sensible "default". Reasoning: It makes sense to compartmentalize each year of WWDC into it's own set of sessions and then within each year, each of the tracks follows a progression (often of introduction sessions, then advanced sessions), so sorting by session number makes sense as well.

In the future I'd love to add different sorting options, so this is just a start.


insidegui:
Great :)

PR #24, opened 2015-05-01
Choose Local Video Storage Directory

donald-pinckney:
Not sure if this is fits with the main goal of the app, but mostly for myself I wanted the ability to change where local videos are stored (external hard drive!).  I implemented it, and it seems to work fine.  Merge if it interests you.


insidegui:
Thanks @donald-pinckney, I ended up implementing this myself.

Issue #21, opened 2015-05-01
Search through ASCII WWDC transcripts in main screen


mac-cain13:
Currently it's only possible to search through the transcript when watching the video. Often I remembers some keywords of a talk, but don't know which one it was about that specific topic. It would be great if the searchbar in the main window with the overview of all videos also searched in the transcripts.

I would imagine that search results are ordered like this:
- hit in the title 
- hit in description
- number of hits in transcript

insidegui:
I don't know, I think this would make the search too complicated.

Currently the transcripts are only downloaded when the transcript window is opened, and the processing of the transcripts is quite expensive, It would be hard to search through all of them on every single keystroke.

That said, if anyone wants to tackle this, go ahead :)

art-divin:
I am almost done with this one.. :wink:

insidegui:
=3

art-divin:
@insidegui I've implemented transcript search, and the keyword is `transcript`. Would be great to add it to README.md...

I had to set -[NSSearchFieldCell setSendsWholeSearchString:] to `YES` because otherwise there would be too much request being sent to `asciiwwdc.com` endpoint.

However, this implies that now users have to press "Enter" in order to start search

insidegui:
As discussed in [this defunct pull request](https://github.com/insidegui/WWDC/pull/26), the best approach will be to cache the transcripts locally and search the local cache.

@art-divin do you want to make this or do you want me to make It? Maybe I can do the cache thing and you can do the search, what do you think?

art-divin:
@insidegui I think it should be separated into 3 aspects:

1. "Fetch Transcripts" Service implementation (you can do that, I've no experience of working with XPC)
(on the second thought, this can be implemented via NSURLSession...). This might include class extension for existing WWDCSessionTranscript object model due to the fact that matches should be counted
2. CoreData - I can do that, basic trick is to "normalize" contents of the transcripts into lowercased indexed strings which will repeat existing class hierarchy
3. UI - this can include number of improvements, including:
 * Moving search method to async mode
 * Async updates of table
 * Sorting according to priorities (query match (year, rank, [...]), title match, description match, number of matches in transcript)

I can see implementation as 3-part one: 

1. Update existing model
2. Implement CoreData
3. Implement Networking
4. Implement UI

UI cannot be done unless everything else is done, so I would say that I'll be able to extend existing model and CoreData, but I need a bit of assistance in Networking.

[...]

PR #26, opened 2015-05-01
Issue 21: search through transcripts

art-divin:
Current implementation doesn't sort according to number/quality of matches, but I think it works relatively good for initial implementation


insidegui:
- The search takes a considerable amount of time
- Multiple words don't work (try searching for "customer sat")
- I tried searching using multiple qualifiers and including "transcript:" but only "transcript:" was considered (try searching "year:2013 transcript:animation")

I really don't think this feature is worth our time, It makes the search too broad and complicated.

art-divin:
@insidegui thanks for the feedback.

Regarding "worthness", I agree that it creates a new format of the search and adds dependency on another endpoint, which doesn't necessarily work fast enough.
However, while developing a lot of iOS 8 stuff during June-December period of 2014, I found myself browsing through WWDC videos just trying to find where exactly did I hear something about some specific new API during actual session in June.

I think that if to fix issue 2&3, the time issue is not so important since this kind of functionality is very useful.

insidegui:
What if we downloaded ASCIWWDC's entire catalog to make a local index? That would make the search a lot faster and maybe allow us to disable "sendsWholeSearchString".

We could download the transcripts for every video we have on the first launch, store them locally (maybe using plists, idk, we may need something like Core Data for this) and every now and then check to see if there are updates.

Later when I add a preferences window (I'll do that soon), we can have a preference to disable transcript search for those who don't want/need It.

art-divin:
@insidegui makes sense, I agree that local search should be implemented even for transcripts.
However, this implies implementing whole other background service for fetching transcripts one-by-one, isn't it?

I think that this PR can be closed without merging, and discussion should be moved in #issue #21

insidegui:
Yup

PR #44, opened 2015-05-09
Allow platform sidebar text color to be inverted on selection

josh-:
Changes the Platform sidebar label to controlTextColor which automatically inverts its color on selection:
screen shot 2015-05-09 at 10 12 45 am


insidegui:
Much better. Thank you!

PR #56, opened 2015-05-27
"Float on Top" option for playback window

dstd:
The feature I'm longing in every video player. Main Menu → Window → Float on Top: Never, While Playing, Always

### DIFF
--- a/WWDC.xcodeproj/project.pbxproj
+++ b/WWDC.xcodeproj/project.pbxproj
@@ -8,6 +8,7 @@
 
 /* Begin PBXBuildFile section */
 		1C967BAA1AE7EB27003662B2 /* NSString+QualifierSearchParser.m in Sources */ = {isa = PBXBuildFile; fileRef = 1C967BA91AE7EB27003662B2 /* NSString+QualifierSearchParser.m */; };
+		7CBE657E1B161B5C0059A52D /* AVPlayerPlaybackStateObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7CBE657D1B161B5C0059A52D /* AVPlayerPlaybackStateObserver.swift */; };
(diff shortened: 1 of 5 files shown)


insidegui:
Thats great, but the code should be reorganized to better match the app's architectural style.

You've implemented this without following the conventions used throughout the app, mainly the way we deal with constants and user defaults. Do you want me to make these changes or do you want to do them yourself?

dstd:
@insidegui sorry, that's my fault - I did a search for 'NSUserDefaults' in the project, but not noticed that the results was inside ASCIIwwdc and DataStore. I hope now I got the conventions correctly :)

insidegui:
Better. There is just a bug: if I close the video window while It's paused, the app crashes.

dstd:
@insidegui yep, `deinit` is not good place to remove observers if they utilize strong refs.
As well as `removeObserver(self)` is not compatible with `addObserverForName`.
Fixed that.

insidegui:
Now it crashes randomly when I close the player window.

dstd:
Could you please show me a stack trace? I can't reproduce any crash now.

insidegui:
captura de tela 2015-05-27 as 11 07 13

That `setRate:` call makes me think It has something to do with the KVO.

dstd:
Yeah, looks like KVO is not best way to observe playback status.

For the reason I can't understand I getting a crash after several rewinds/forwards using a keyboard.
With zombies I found an error message - `[CFNumber retain]: message sent to deallocated instance`.

After some trials I ended with saving the value of `change[NSKeyValueChangeNewKey]` into an instance variable, which solve a crashes. But this is definitely a smelly solution, so I will rewrite this code.

insidegui:
Maybe use notifications? I think there are notifications to know when the playback has started. Or use a time observer...

dstd:
@insidegui, time observer is not suitable, because according to docs it reports only during a playback.
Unfortunately I haven't found any events/delegates/notifications for the playback states, the only way suggested on StackOverflow is to use KVO.
I've replaced KVO with NSTimer-based polling class. Ticking once a second seems to be not too power consuming.

insidegui:
I'll see if I can improve It a little, thanks for all your work :)

Issue #51, opened 2015-05-15
When a video is downloaded there's no built-in way to delete it


dmthomas:
When a video is downloaded locally, perhaps the download button can become a delete button so that the video can be removed without browsing to the location manually to delete the video.

insidegui:
Great idea!

insidegui:
Implemented by @finngaida in #119

PR #58, opened 2015-06-03
Added support for deleting a locally cached video #51

amayers:
Needs an icon instead of the text 'X' that it's using now on the button.


insidegui:
Please add an alert to ask the user to confirm before deleting the video, It's too easy just to click on the button and "boom", the video is gone. I'll take care of the icon. Thank you!

amayers:
Will do.

PR #59, opened 2015-06-03
Fix crash when computer sleeps

amayers:
When the computer sleeps in the middle of a checkForLiveEvent, request data comes back empty, but non-nil.


insidegui:
Thanks!

Issue #54, opened 2015-05-19
Scroll Indicator on Transcript Window Goes Beyond Search Field


dmthomas:
The scroll gutter on the transcript window extends underneath the search field pane on the bottom of the window.

PR #61, opened 2015-06-05
Inset transcript web view so it doesn't hide behind the effect view -…

TimothyChilvers:
… fix for #54


insidegui:
Thanks!

TimothyChilvers:
NP. There's a one line extra space at the bottom of the content, but I couldn't quickly see what was causing it.

PR #71, opened 2015-06-10
Localize the date/time used in the live banner

kballard:
This was primarily motivated by a desire to stop seeing 24-hour time.


insidegui:
Thanks!

PR #78, opened 2015-06-12
New search qualifier -- favorited

mcgraw:
Needed a way to clear noise & reveal sessions I had favorited.

i.e.
year:2015 favorited:yes


insidegui:
The logic was wrong (It assumed `favorited:yes`). Fixed and merged. Thanks!

PR #88, opened 2015-06-18
Video window float on top

jagreenwood:
Added menu item and logic to support keeping the video window on top of other windows.


insidegui:
Thanks!

PR #89, opened 2015-06-26
Search using keywords.

erndev:
Search using keywords instead of the literal text typed in the search field.

This allows searching sessions even if you don't know the exact session title.

For instance, typing "new cocoa" would find al sessions like "What's new in cocoa..."  without the need to type the exact session name. And you could also fine tune that search to "new cocoa touch" to focus even more the search.


insidegui:
Cool!

PR #102, opened 2015-11-01
Adds skipping forward and back tip

philosopherdog:


insidegui:
Swiping on a Magic Mouse works as well, can you please include It?

PR #104, opened 2015-12-16
fix filter bar's wrong frame when use mouse

wangmchn:
Hi, @insidegui 
Thx for the nice repo, I enjoyed a lot and learned a lot.
When I use it (with a mouse), I find the `filter bar` will have a wrong vision when scroll. So I try to fix.
Finally I find out that when I use mouse wheel, the `theEvent.phase` and `theEvent.momentumPhase` is equal to `NSEventPhaseNone`, so it never call the method `-scrollViewDidEndDragging:`, i fix it as the pr.
Maybe it's not the most elegant way to deal with, just give an idea.
Thanks again!


insidegui:
Works great. Thank you!

PR #119, opened 2016-03-31
add 'remove download' option

finngaida:
I found that the app lacks an option to conveniently delete downloaded session videos without having to dig through `~/Library`, so I made these small changes.

Unfortunately the pull request is a little cluttered by the new Realm changes, but as I had updated to Xcode 7.3 before, the Embedded Framework wouldn't compile anymore.
So in a way this is also a Swift 2.2 compatibility commit 😉 

Also changed the screenshot references in the readme to relative paths.


finngaida:
Also I found that in order to compile I had to clean the project beforehand.

insidegui:
Thanks @finngaida! I will check It out and merge ASAP 😁

insidegui:
There's a weird crash going on.

Steps to reproduce:
- Filter the videos list to show only  downloaded videos
- Select one or more videos and choose "remove from downloads"
- Select the last video on the list
CRASH

Can you take a look at It?

crash

finngaida:
Hmm, somehow I can't resemble the crash by any means...
But as the log shows the Realm db seems to call `objectAtIndex:` and then runs out of bounds, so I updated the `removeDownload` method to also update the dowloaded status in the database.

I'm not sure about the data source model of the table view, because I seem to not be able to actually remove the deleted session entry from the list of downloaded items, although the little downloaded symbol as well as the download button in the top right seem to change their state appropriately, maybe you can give me a hint on where to look to fix that and also check if this recent commit fixed the crash for you.

insidegui:
There is a little issue with the way downloads are stored. Initially there was no flag on the database to determine whether an item was downloaded or not but when I was implementing the new search I had to add It. The button and the icon (I think) are updated based on the existence of the file on disk, if you delete the file through finder they update as well, but I'm not sure if the flag on the database is automatically updated.

insidegui:
Thinking a little bit more, I think the problem is happening because the list is not updated after the download is removed. Since I'm filtering by "downloaded", if I remove the download that session should disappear from the list. I think I can fix It ;)

insidegui:
I was able to fix the issue by reloading the table after a video download is removed.

finngaida:
Awesome! And thanks for the support!

PR #131, opened 2016-05-25
Temporarily disable video controls while loading

peterkos:
This prevents a crash when a user attempts to click on the time slider when the video is still loading by hiding the controls while the video is loading.


insidegui:
I hadn't seen this, I would've merged It earlier if I had :)

Thanks!

Issue #136, opened 2016-06-10
Inconstancy with downloads from pre 2013


SoneeJohn:
Using the latest release of the application (4.1.2 at the time of writing) the following happens:

- Right click on a session from pre 2013 (e.g. 2011) and the contextual menu has a download option.
 - However, the right split view, that has additional details does not contain a download button.

I uploaded screenshots to demo this. Note videos from pre 2013 cannot be downloaded therefore, having the download option in the contextual menu seems like an inconstancy IMO. Not a huge issue but, I was hoping to bring it to your attention.

![screen shot 2016-06-09 at 8 21 27 pm](https://cloud.githubusercontent.com/assets/4827301/15950812/4ac90a8c-2e80-11e6-9e11-8f00dd9faa0f.png)
<img width="467" alt="screen shot 2016-06-09 at 8 22 50 pm" src="https://cloud.githubusercontent.com/assets/4827301/15950813/4bd9c984-2e80-11e6-83dc-12ec0d3357f3.png">

insidegui:
I am aware of this, It is a usability issue and will definitely be fixed. Thanks for reporting.

SoneeJohn:
@insidegui I think if fixed this with this PR: #137

PR #137, opened 2016-06-10
Improve constancy between the tableView's NSMenu and the VideoDetailsViewController

SoneeJohn:
- Videos from 2013 and above will be evaluated
- if a session isScheduled then both the `downloadMenuItem` and
`removeDownloadMenuItem` will be disabled

else the menu items are evaluated based on whether or not it has been
downloaded
- any other session (pre-2013 ) the menu items are disabled

Note: I haven’t throughly tested this but, it works fine so far.

#136


insidegui:
Thanks! Can you please remind me to merge this after next week? I won't me merging any PRs until after WWDC because any change to the app I'm making right now is being very thoroughly tested in various circumstances to make sure everyone can watch the sessions live without major issues :)

SoneeJohn:
@insidegui Sure no problem 😊

bre7:
Hi @SoneeJohn, I noticed the same things you did and wanted to add some details to your PR. Would you mind adding the same eval for "Mark as Watched"/"Mark as Unwatched" and "Add to Favorites"/"Remove from Favorites" ?

insidegui:
Can you please make the following changes before I merge:

- Move the `NSMenuDelegate` conformance to an extension instead of declaring It directly on the view controller class
- Use `guard` and early return instead of nested if statements
- Not connect menu items directly to the controller with outlets, give each menu item a tag and validate based on the tag (you can use an enum to encapsulate the meaning of each tag:

```
    private enum MenuItemTags: Int {
        case Download = 1001
        case RemoveDownload = 1002
    }
```

SoneeJohn:
@insidegui Quick question because I'm new to Swift. If I move `NSMenuDelegate` from the view controller into an extension how will assign the `tableView`'s `NSMenu`'s delegate so that I can be notified when the menu is about to open? I was thinking instead to do this: 

```
override func validateMenuItem(menuItem: NSMenuItem) -> Bool{
        
//Validate the items
        return false
        
    }
```

SoneeJohn:
Hi @bre7 I made the changes to validate those items as well.

insidegui:
You can just declare an extension on `VideosViewController ` which conforms to `NSMenuDelegate` and implement the delegate methods in there. The assignment to the table view can still be done on the main class.

```swift
extension VideosViewController: NSMenuDelegate {
    // Implement NSMenuDelegate methods here
}
```

You can also move your enum into this extension to make the code a little cleaner ;)

SoneeJohn:
@insidegui Thanks :). I ended up using `func validateMenuItem(menuItem: NSMenuItem) -> Bool`, take a look and tell me what you think.

PR #160, opened 2016-06-19
Move realm framework manually into project

SoneeJohn:
I was unable to build the project by following the build instructions in the `README`. I believe it would be beneficial to contributors if the Realm framework was manually included into the project and not integrated via CocoaPods. When using CocoaPods I got the following error: `Could not build Objective-C module 'Realm'`. 
screen shot 2016-06-19 at 11 52 37 am

You can also integrate using Carthage but, that downloads the framework for all platforms which makes the project a little too large IMO.

**Summary of charges**:
- Remove 'pod 'RealmSwift' from Podfile
- Add RealmSwift and Realm directly in project.


insidegui:
I'm not moving the dependencies back into the project, that's not a good practice. You probably have something wrong with your setup. Can you open an issue describing this problem? Then we can discuss It in there?

Thanks!

SoneeJohn:
@insidegui sure no problem.

SoneeJohn:
@insidegui reported issue here #161.

PR #170, opened 2016-06-27
Feature: Hidden Preference for Realm Database Path

tternes:
I noticed in the roadmap there are plans for synchronizing data between devices in the future. Until then, I've put together a workaround to store the Realm database inside an arbitrary folder by writing a hidden preference:

`defaults write br.com.guilhermerambo.WWDC realmDatabasePath "/Users/thaddeus/Dropbox/WWDCAppDatabase"`

There are likely quite a few issues with this (especially with Dropbox), but it works to serve as a backup of my configuration at present. I've started working on UI updates to the preferences window to surface this feature, but I feel that only further exposes the limitations, so I've stopped working on that for now.

Given the limitations of this functionality, I'm not going to be offended by it not being merged. However, if it's something that might be useful to others, it could probably be cleaned up a bit more and treated as an undocumented/hidden feature.

### DIFF
--- a/WWDC/Preferences.swift
+++ b/WWDC/Preferences.swift
@@ -35,6 +35,7 @@ class Preferences {
         static let automaticRefreshEnabled = "automaticRefreshEnabled"
 		static let floatOnTopEnabled = "floatOnTopEnabled"
         static let automaticRefreshSuggestionPresentedAt = "automaticRefreshSuggestionPresentedAt"
+        static let realmDatabasePath = "realmDatabasePath"
         
         struct transcript {
             static let font = "transcript.font"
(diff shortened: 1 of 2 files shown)


insidegui:
I don't think this is a good idea. I thought about doing something like this before but it sounds really dangerous and the app is not engineered to take into account a database that might change from under it. I think we'll leave this to the next release, where user data will be saved in iCloud instead of Realm :)

tternes:
> the app is not engineered to take into account a database that might change from under it. 

Agreed - having the UI available to change the preference highlighted this shortcoming dramatically.

> data will be saved in iCloud instead of Realm

👍

PR #212, opened 2017-06-02
Added 1.25 speed for player

azimin:
Hi @insidegui, I totally love this great app and used it a year ago!
I love this speed feature now, but sometimes I'm missing 1.25 (when 1.0 is slow, 1.5 is fast).

If I'm not the only one who is missing it, maybe we can add?
Thanks!


insidegui:
Thanks 🎉

Issue #209, opened 2017-06-02
Could I please ask how to cancel a downloading task?


samingzhong:

insidegui:
This is not implemented yet.

samingzhong:
Come on guy!👍

insidegui:
Implemented in 5.0.6. Thanks @CassiusPacheco

PR #214, opened 2017-06-04
Add support to cancelling download by clicking on the progress indicator

CassiusPacheco:
That's a quick fix for the issue https://github.com/insidegui/WWDC/issues/209.

The concept is based on the iOS App Store's download handling system. I'm not 100% sure if it can be applied for macOS too.

If so, ideally we'd have a progress indicator with a square (cancel symbol) in the middle, like the iOS App Store does, so the user is aware of the feature.

However, I think it's still worth shipping it as a quick fix anyways 😊 :shipit: 

Thanks for the awesome project @insidegui, I love it <3

### DIFF
--- a/WWDC/AppCoordinator+SessionActions.swift
+++ b/WWDC/AppCoordinator+SessionActions.swift
@@ -14,6 +14,14 @@ import PlayerUI
 
 extension AppCoordinator: SessionActionsViewControllerDelegate {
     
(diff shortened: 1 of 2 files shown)


samingzhong:
Great job!

insidegui:
Well done @CassiusPacheco ! 🎉

PR #219, opened 2017-06-04
Removed unused variable assignment from PUIPlayerView

ezfe:
Firstly I want to preface this by saying that it's not clear why this line was included in the first place and there may be a reason it exists. That being said, it doesn't appear to serve any purpose and the player runs without it.

An alternate solution to satisfy the compiler warning would be

```swift
_ = playerItem.currentMediaSelection
```


insidegui:
I refactored and forgot to delete it. Tks!

PR #229, opened 2017-06-07
Calendar Support

SoneeJohn:
I noticed today while I was viewing sessions on my computer that there is no support for quickly adding a session to the calendar. The officially WWDC app has this so I thought I should add it. 

By the way, thanks for the awesome app @insidegui 🙂

TODO:

- [x] Support for adding session to the calendar

- [x]  Only show calendar button when WWDC is active
 
- [x]  Store eventIdentifier somewhere so it can be referenced later so the user can remove the event from the app


insidegui:
Thanks @SoneeJohn ! I'll review this as soon as I can

SoneeJohn:
@insidegui I was planning to save the session identifier along with the eventIdentifier of EKEvent so that the user can remove it from their calendar. However, I was wondering if that would be too much.

insidegui:
@SoneeJohn you can add that as a property to `SessionInstance`. Don't forget to increase the `coreSchemaVersion` in `Constants.swift` so the database gets migrated correctly. If you don't know how to work with Realm, you can find the documentation here: https://realm.io/docs/swift/latest/

SoneeJohn:
@insidegui thanks! I'll look into this tomorrow.

insidegui:
@VicenteBorrell Can you get us an icon for the "add to calendar" feature? It will only be useful next year, but nice to have :)

VicenteBorrell:
Yeah, no problem! I was thinking, maybe we could leave it for next year. I would like to improve the session detail layout, adding location and time, and put the icon next to it.

SoneeJohn:
@insidegui finished :-) All what might be needed now is a new calendar icon from @VicenteBorrell

insidegui:
@VicenteBorrell Since @SoneeJohn has already done all the hard work of implementing calendar support, I don't think we should wait until next year to merge this. Can you get us an icon for the feature?

Thanks

VicenteBorrell:
You can find the icon on the Sketch file @insidegui 
Here it is too:
calendar icon 2x

SoneeJohn:
I updated the icon. Thanks @VicenteBorrell !

insidegui:
@SoneeJohn I made some changes to the implementation and migrated it to our code style and recent changes. You can see the updated branch here: https://github.com/insidegui/WWDC/tree/calendar

Can you make it so the calendar button only appears for sessions that are in the future? There's no point in showing it for sessions that have already passed.

SoneeJohn:
@insidegui I see @bcmn beat me to it ;-) https://github.com/insidegui/WWDC/commit/5c14187ea8b36228d7810b0ad3655248f9311394

bcmn:
Oh oops, I thought I’d commented here, sorry. 😬 Gui tried to prod me too after I added it just from seeing activitu on the branch 😂 Yeah, it’s currently based on the endTime of the session, which seems like it allows a bit more buffer for edge use cases, I don’t know if anyone disagrees.

SoneeJohn:
@bcmn No worries 😂. I think it would be better if the button simply don’t show up if the startTime is now in the past.

bcmn:
Yeah, I think you're right. I think I wavered when deciding and then made the choice based on an incorrect memory of how much of the event details we surface in UI. I changed it.

insidegui:
Yeah I agree, there's no point in adding an event that has already started to your calendar ¯\_(ツ)_/¯ 

I'll do some more testing tonight and merge after reviewing the changes.

Thanks

PR #242, opened 2017-06-09
Forward/backward 15s buttons

sfsam:
I added forward and backwards 15s buttons because that is often a useful time increment when reviewing WWDC material. It is the same increment the official iOS apps use. Also, the buttons show 15 or 30 to visually indicate how they behave.


insidegui:
Can you provide a screenshot of how they look on the interface?

sfsam:
Here's a screenshot:
screen shot 2017-06-09 at 1 49 42 pm

insidegui:
I see. Since this changes the design of the buttons and the player, we need @VicenteBorrell's approval.  

I personally think having four buttons like that is not the best solution. Maybe we could have a preference so the user can set how much time they skip/go back by default and allow the alternate to be used by clicking and holding or control clicking the button.

@VicenteBorrell what do you think?

VicenteBorrell:
Four buttons is too much. But I like the idea of the icons, although with two arrows (I know the current ones don't but it's on my todo list)
I normally like 15s backwards and 30s forward, but that's not symmetric XD
Either that or both 30s (the safest option).

sfsam:
Ok, thank you for considering it.

FWIW, I had considered either a preference or alternate method like right-click or double-click, but ended up with 2 more buttons because it seemed the best experience when using it. The options are immediately obvious with no configuration and nothing hidden.

It's a great app, thanks for making and sharing it!

allenhumphreys:
I wouldn't mind an option!

insidegui:
@VicenteBorrell @sfsam I'm leaning towards having a preference or a second option with control click. I don't feel the need to skip 15s OR 30s, having only one of the options is good enough for me. If we were to choose one of them, I think 15s is better because you can easily skip 30s by just pressing the button twice, while if we leave just 30s, you can't skip less time.

sfsam:
@insidegui @VicenteBorrell I agree 15s is a better default than 30s.

insidegui:
@sfsam Can you change your implementation so it shows the 15s or the 30s base on a preference?

If you don't want to add the preference to the prefs window you can just add the property to `Preferences.swift` and I'll add the switch to the window later 🙂

Thanks!

insidegui:
Since the preference will be implemented in the app and not in `PlayerUI`, I think you should add a method to ` PUIPlayerViewAppearanceDelegate` which determines the button to show (15s or 30s).

sfsam:
@insidegui I've added a switch and code to change the preference from a default of 15 seconds to 30 seconds. Here is a screenshot:

wwdc skip preference

insidegui:
Great work @sfsam! Looks good to me. @VicenteBorrell do you want to change the glyphs or can I merge this?

VicenteBorrell:
Looks good to me too :) @insidegui

PR #243, opened 2017-06-09
Sets audio algorithm to time domain

allenhumphreys:
I noticed that while playing back the session at a rate higher/lower than 1.0, the audio didn't sound so great. This change switches to the time domain algorithm which is more suited for these primarily voice-only videos.


insidegui:
Good job. Thanks!

PR #264, opened 2017-06-27
Add playback speed menu when right-clicking the player button

pedrovieira7:
This PR adds a player playback speed menu when the user right-clicks on its button. I personally did this out of necessity, so I hope more people find it useful too.

Here's how it looks like:
captura de ecra 2017-06-27 as 04 02 59


bcmn:
Great stuff, thanks very much! 😊

Issue #230, opened 2017-06-07
Add useful menu items, with associated keyboard shortcuts


unnamedd:
Shortcut suggestions
- [ ] Mark as watched
- [ ] Mark as unwatched
- [ ] Mark as favorite
- [ ] Next session
- [ ] Previous session
- [ ] Go to Filter
- [ ] Next Tab (Schedule | Videos)
- [ ] Previous Tab (Schedule | Videos)

These are suggestions that I can see as useful for now.

Issue #227, opened 2017-06-06
Searching prevents spaces


DevanB:
I was perusing the Schedule tab today and needed to search for "Design Studio Shorts". To my surprise, I found that the search is preventing spaces. I'm assuming this is a bug, but I could be wrong. I wish I knew more about macOS app development because I'd love to detail where the problem may be and submit a PR.

Thanks for the app!

insidegui:
This happens when the player is active because it "swallows" all space key presses 😬

Will fix. Thanks

bcmn:
Seemingly fixed in #387 & #478.

PR #269, opened 2017-07-01
Add arrow shortcuts to go back/forward

pedrovieira7:
This PR adds the ability to go back/forward in time on the player using the left and right arrows respectively. It also respects the user preference of 15s/30s when executing the shortcut.


bcmn:
There's currently some conflicts with different sections of the app consuming key commands in inconsistent and conflicting ways (Such as #227). Some hacky code may have found its way in in the pre-WWDC panic. 😜😅 @insidegui contacted me to say he thinks it's best if we restrain from adding any further commands right now, and we'll instead include similar functionality to this PR in #230, which we'll now treat as a way of tracking this larger effort to update menu items and associate key commands with those.

Sorry about this!

pedrovieira7:
@bcmn no worries, thanks for the heads up!

PR #273, opened 2017-07-03
Fixes PIP closes while filters are active

allenhumphreys:
The app behaves weirdly if you do the following:

1. Start playback
2. Enter PIP mode
3. Apply a filter to your list that excludes the currently playing session
4. Exit PIP mode

I fixed the issue by simply not filtering the current session, another fix would be from the player side by handling the case where the session isn't in the list and not allow PIP to exit or continue playback offscreen?

### DIFF
--- a/WWDC/SearchCoordinator.swift
+++ b/WWDC/SearchCoordinator.swift
@@ -158,7 +158,14 @@ final class SearchCoordinator {
             subpredicates.append(Session.videoPredicate)
         }
         
-        let predicate = NSCompoundPredicate(andPredicateWithSubpredicates: subpredicates)
+        var predicate = NSCompoundPredicate(andPredicateWithSubpredicates: subpredicates)
+
+        if let appDelegate = NSApplication.shared().delegate as? AppDelegate,
+            let currentlyPlayingSession = appDelegate.coordinator.currentPlayerController?.sessionViewModel.session {
+
+            // Keep the currently playing video in the list to ensure PIP can re-select it if needed
+            predicate = NSCompoundPredicate(orPredicateWithSubpredicates: [predicate, NSPredicate(format: "identifier == %@", currentlyPlayingSession.identifier)])
+        }
         
         #if DEBUG
             print(predicate)


insidegui:
We thought about this issue when working on the design for the auto-PiP functionality.

I think if we can't find a better solution we should go with this one since I've had this issue myself and it's really annoying 😅

@bcmn @VicenteBorrell Any thoughts?

allenhumphreys:
I also considered the following:

1. When the player is exiting PIP, during the process of selecting the playing video, see if it's in the list and if it's not, force it into the list. Complexity is quite high and would probably be a nightmare to maintain.

2. Offscreen playback

Additionally, we could improve the current implementation by triggering a filter update when the currently playing session changes.

insidegui:
I think we thought about having a section at the bottom of the list dedicated to the "now playing" session (similar to iTunes). Not sure about it tho.

allenhumphreys:
Not a bad idea at all. Still more complex than this solution!

allenhumphreys:
We could take this fix for now and create another task to do something more "interesting".

VicenteBorrell:
@allenhumphreys Not filtering the current session is not a valid solution, and introduces other problems.

I think it should either show to the video detail without selecting it on the sidebar (as it isn't there), or keep what's already on that view.

cc @insidegui @bcmn

allenhumphreys:
Well, I would say the validity is rather up for interpretation then.

insidegui:
Thanks for your feedback @VicenteBorrell.

I'm actually going to merge this for now because I think it provides at least a temporary solution for a real issue. It might not be the best solution for the long term but it's better than what we have now.

allenhumphreys:
@insidegui Thanks. Now that this is merged, would you like me to look into @VicenteBorrell's suggestion of showing the session details while not relying on the ability to select the session in the table? I've noticed some odd behavior around selections, specifically how selections are handled across filtering operations. I could look into improving that at the same time.

bcmn:
I fear trying to decouple the relationship between sidebar selection & what's shown in the detail view would introduce a lot of complexity.

allenhumphreys:
@bcmn Strictly speaking it wouldn't actually decouple the selection. It would override it temporarily. Though, like I said before, there are already some problems with the selection logic. But a very basic override would like like [this](https://github.com/insidegui/WWDC/compare/master...allenhumphreys:alternative-solution)

VicenteBorrell:
@allenhumphreys I said it isn't valid because the results of the filtering in that case aren't valid :)

@bcmn @allenhumphreys I don't know about the complexity but the reasoning is that the sessions is selected, but because it's filtered, it's not shown.

allenhumphreys:
It seems like things are pointing toward having a now playing section of some kind. It would both 1. Not decouple the selection 2. Not mess with the expected results of the filters.

Issue #282, opened 2017-08-12
Context menu acts on a wrong selection


antonselyanin:
- Version of macOS
10.12.6

- Version of the app
5.1.3

- What were you trying to do when the issue happened
While I was watching a video I tried to add other video to favorites

- If it is a wrong behavior, list all steps required to reproduce it
(1) Select a video from the list, 
(2) Right click any other video - context menu appears
(3) The context menu acts on the primary selection while the standard OS behavior is to act on the secondary selection.

I checked the OS behavior in Finder app :)

PR #283, opened 2017-08-12
Fixed context menu behavior (act on secondary selection)

antonselyanin:
Fix issue #282


insidegui:
Thanks for your collaboration.

We experienced a crash when trying to act on multiple selections. The crash was in `SessionsTableViewController` line `377` and it was an `EXC_BADACCESS` crash. It is not always reproducible.

antonselyanin:
@insidegui 
I managed to reproduce it, you just need to click on an empty area of the table view (start searching for something unique)

insidegui:
Cool! Thanks

PR #287, opened 2017-08-15
Use system provided table filtering for session list

allenhumphreys:
So, this is a thing I was playing with. It adds animations to the filtering operations and leverages the system selection to provide sensible selection preservation.

giphy giphy-2

Looking for feedback

### DIFF
--- a/WWDC/MultipleChoiceFilter.swift
+++ b/WWDC/MultipleChoiceFilter.swift
@@ -7,7 +7,6 @@
 //
 
 import Foundation
-import SwiftyJSON
 
 struct FilterOption: Equatable {
     let title: String
--- a/WWDC/SessionDetailsViewController.swift
+++ b/WWDC/SessionDetailsViewController.swift
@@ -18,29 +18,42 @@ class SessionDetailsViewController: NSViewController {
     
     var viewModel: SessionViewModel? = nil {
         didSet {
-            menuButtonsContainer.isHidden = (viewModel == nil)
-            
-            self.shelfController.viewModel = viewModel
-            self.summaryController.viewModel = viewModel
-            
-            if viewModel?.identifier != oldValue?.identifier {
-                self.transcriptController.viewModel = viewModel
-                
+
+            informationStackView.animator().isHidden = (viewModel == nil)
+            shelfController.view.animator().isHidden = (viewModel == nil)
+
+            shelfController.viewModel = viewModel
+            summaryController.viewModel = viewModel
+            transcriptController.viewModel = viewModel
+
+            guard let viewModel = viewModel else {
+                return
+            }
+
+            if viewModel.identifier != oldValue?.identifier {
+
                 showOverview()
             }
             
-            self.transcriptButton.isHidden = (viewModel?.session.transcript() == nil)
+            transcriptButton.isHidden = (viewModel.session.transcript() == nil)
             
-            let shouldHideButtonsBar = self.transcriptButton.isHidden && self.bookmarksButton.isHidden
+            let shouldHideButtonsBar = transcriptButton.isHidden && bookmarksButton.isHidden
             self.menuButtonsContainer.isHidden = shouldHideButtonsBar
-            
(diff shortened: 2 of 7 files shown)


insidegui:
This looks amazing. @bcmn can you test and merge please? :)

bcmn:
There's a strange UX issue I've encountered with this -- after selecting a video & enabling filters that exclude said selected video, it becomes impossible to play the video that is shown in the main, right-hand panel. I don't know if this was intentional or whether it's something that slipped through.
For example: Select a 2017 video, then filter to 2016 only, & immediately try to play the video in the right-hand panel.

The current UX for this situation is that the video shown in the right-hand panel is changed to a video that is included under the currently active filters. I think there's an argument that this could be a little frustrating or odd, but I think this new behaviour is _more_ odd -- & would arguably even look like a bug or issue to users.

allenhumphreys:
Thanks for looking at it. That's a good catch. What prompted this initial work was that selections used to be preserved by index across filtering, which meant that if you selected a video, then filtered that video out, a random video would become selected. I believe you can see that behavior in the current production version. 

The case that a video is already playing is already handled since it won't be filtered, but should the behavior in this case be that the first item in the filtered list become "selected" for display in the shelf? (While not actually affecting selected, i.e. highlighted table items) I think this would be similar to a fresh app state?

allenhumphreys:
Also, it was definitely something I missed.

bcmn:
> should the behavior in this case be that the first item in the filtered list become "selected" for display in the shelf? (While not actually affecting selected, i.e. highlighted table items) I think this would be similar to a fresh app state?

Playing around with the two behaviours, I think this would be a good choice (though currently resetting to the first cell index would constantly revert to the currently playing video, which might become irritating). I'm curious whether @VicenteBorrell or @insidegui would have any thoughts?

allenhumphreys:
I could definitely try to fix the reason the video in the shelf isn't actually playable, if you'd rather go that route?

insidegui:
> I could definitely try to fix the reason the video in the shelf isn't actually playable, if you'd rather go that route?
I believe that would complicate things quite a bit because we would have to break an assumption I've made in several places (that the currently selected video on the table view is always in sync with the selection variable in the coordinator).

allenhumphreys:
Hey @insidegui @bcmn, quick update. I don't have a fix for the issue where you can't play a video after it's been filtered, but it turns out that this issue wasn't caused by me. You can reproduce the issue in master by command + click to deselect the current selection. The `SessionDetailsViewController` doesn't correctly apply a null state to itself when the `viewModel` gets set to nil.

Currently, this is what happens:
giphy-3

I could prevent setting the `viewModel` to nil:
```
didSet {
            if viewModel == nil {
                viewModel = oldValue
            }
```

Or we can properly handle the null state by clearing the entire details page.

***Insert Work Here***

The first option is fastest. From an architectural and UX perspective, the second option is probably best. But there should be a message in the details screen when in the null state to the effect of 
"<----- choose a session, dummy" 
(Obviously, I'm not a UI designer)

insidegui:
That's interesting. I think we should just disallow empty selections, we had an empty state in v4 but I don't think it's particularly useful.

insidegui:
@allenhumphreys Can you update this so it's impossible to have an empty selection in the table view? I really want to merge this and release it with the new version we'll put out this week :)

allenhumphreys:
I won't be able to get to that until sometime after the weekend. I'm going to the eclipse on Monday and work has been pretty busy :) but it's definitely on my list.

VicenteBorrell:
Hi! Just saw this :)

I vote for having the first session of the filtered results selected on the sidebar and update the session detail. But if the selected session is included in the results, it should stay selected, right?

Also, if there are no results, no session detail should be shown (not currently implemented).

allenhumphreys:
@vincenteborrell Good point on the no results. That means I'll definitely need fix that null state issue. I'll have it done within a week. I'll try to do it Wednesday or Thursday.

VicenteBorrell:
Thanks @allenhumphreys ! Enjoy the eclipse! :)

allenhumphreys:
I'll reopen when it's ready.

PR #298, opened 2017-08-31
Avoid over iterating when filtering

otaviolima:
Use array.first(where:) instead of array.filter().first
Use array.contains(where:) instead filtering and checking for filtered array size
Use array.reversed().first(where:) instead of array.filter().last

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -125,19 +125,11 @@ public final class PUIPlayerView: NSView {
     }
 
     public var firstAnnotationBeforeCurrentTime: PUITimelineAnnotation? {
-        return annotations.filter({ annotation in
-            guard annotation.isValid else { return false }
-
-            return annotation.timestamp + 1 < currentTimestamp
-        }).last
+        return annotations.reversed().first { $0.isValid && $0.timestamp + 1 < currentTimestamp }
     }
 
     public var firstAnnotationAfterCurrentTime: PUITimelineAnnotation? {
-        return annotations.filter({ annotation in
-            guard annotation.isValid else { return false }
-
-            return annotation.timestamp > currentTimestamp + 1
-        }).first
+        return annotations.first { $0.isValid && $0.timestamp > currentTimestamp + 1 }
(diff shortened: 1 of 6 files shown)


ConfCore/Session.swift
```diff
@@ -67,7 +67,7 @@ public class Session: Object {
         guard let realm = self.realm else { return nil }
         guard !transcriptIdentifier.isEmpty else { return nil }
         
-        return realm.objects(Transcript.self).filter("identifier == %@", self.transcriptIdentifier).first
```
allenhumphreys: This `filter` method is a realm query not a filter loop. I'm pretty sure it gets turned into SQLLite query.

The first code block from [the docs](https://realm.io/docs/swift/latest/) includes an example `let theDog = realm.objects(Dog.self).filter("age == 1").first`

Though, from a Swift perspective, this is generally the correct move.

For an explanation of why [see this](https://realm.io/docs/swift/2.10.0/api/Classes/Results.html#/s:FC10RealmSwift7Results6filterFtSSGSaP___GS0_x_)
>`Results` are lazily evaluated the first time they are accessed; they only run queries when the result of the query is requested. This means that chaining several temporary `Results` to sort and filter your data does not perform any unnecessary work processing the intermediate state.

insidegui: @otaviolima Did you perform any measurements to confirm that your changes actually make things faster?

@allenhumphreys Realm is not SQL-based, but from my understanding, when using predicate formats for filtering it is more performant. I might be wrong tho, maybe we can ask someone from Realm to clear things up regarding their filtering methods.

insidegui: @otaviolima BTW, I'm only talking about the places where predicate formats got replaced by closures, most of the changes are clearly for the better ;)

allenhumphreys:
@insidegui haha, you're right about the tech, but I still believe in the general point I was making 😬

otaviolima:
@insidegui As @allenhumphreys pointed out the Realm code should remain as it it.

I tested and the performance drop is huge.

I removed all the changes that were related to Realm, rebased and pushed again.

PR #299, opened 2017-09-01
Adam Nemecek's refactor

allenhumphreys:
This is meant to replace #241 [abbreviated diff](https://github.com/insidegui/WWDC/compare/b527de4a7b677d663b806becf2737028f550e1af...adamnemecek:master?w=1#diff-fb0953e87681a1a236701302d365ee56)

To show the diff without whitespace-only changes, use [this](https://github.com/insidegui/WWDC/pull/299/files?w=1)

This is just a complicated rebasing of @adamnemecek 's PR because I can't stand to see it left open.

Also, regarding the trailing white spaces, everyone knows there is an Xcode setting to trim trailing white spaces even on empty lines, right? Highly recommend turning it on. I don't know why it's not on by default.

PR #300, opened 2017-09-03
Fix notification handling bug introduced by pull #299

otaviolima:
By this diff: https://github.com/insidegui/WWDC/pull/299/files?w=1#diff-7b65a819f91f73ac1b15377c72ca09dcR277 we see that the code was wrongly refactored.


insidegui:
Very well done! I was against this change in the original pull because the reduced size of the code made it harder to understand and could lead to bugs. Guess I was right 😬

insidegui:
Ops, closed by mistake.

PR #308, opened 2017-09-06
Fix #307 by improving equality checking of FilterOption

### LINKED ISSUE #307: Schedule "events" filter preservation doesn't distinguish between the 2 in the UI
- Version of the app
built from master
- What were you trying to do when the issue happened
reproduce #306 

1. Filter by Session
2. Quit the app gracefully to ensure filters are saved
3. Reopen the app, open the popup for Events

Expected results:
"Sessions" is checked

Actual results:
Both values are checked

allenhumphreys:
The equality checking function is used by `contains`. Specifically, the problem was manifested in `SearchFiltersViewController` line 238 when determining whether to place a checkmark next to a filter.

### DIFF
--- a/WWDC/MultipleChoiceFilter.swift
+++ b/WWDC/MultipleChoiceFilter.swift
@@ -24,7 +24,7 @@ struct FilterOption: Equatable {
     }
 
     static func ==(lhs: FilterOption, rhs: FilterOption) -> Bool {
-        return lhs.value == rhs.value && lhs.title == rhs.title && lhs.isNegative == rhs.isNegative
+        return lhs.value == rhs.value && lhs.isNegative == rhs.isNegative && lhs.title == rhs.title
     }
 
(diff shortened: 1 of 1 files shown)


allenhumphreys:
@otaviolima Would you care to double check my logic since you have a knack for it? 😄

insidegui:
Waiting on feedback from @otaviolima to merge 😊

otaviolima:
Looks good to me. 

If I was to make any suggestion would be putting the `Bool` comparison as first due it is faster than `String` comparison and could shortcut the whole expression.

----
That is one good example of using [Sourcery](https://github.com/krzysztofzablocki/Sourcery) to generate equatable protocol conformance, we could investigate if it is worth to use it on the project.

allenhumphreys:
I left the value comparison as the first one since it is the one that is most likely to be different and lead to a short circuit.

otaviolima:
That makes sense!

Issue #306, opened 2017-09-06
Session list disappears


otaviolima:
Using 5.1.3

1. Do a search
2. Close the app
3. Open back
4. Clear search

Expected behavior:
List of sessions 

Current behavior:
Empty list.

Workaround:
Toggling filter button makes the list appear again.

insidegui:
Is this in the current release version or master?

otaviolima:
I tested on release, once I get home I will test against master.

otaviolima:
Can't reproduce using master.

Issue #305, opened 2017-09-05
Unintended scroll when switching tabs 


insidegui:
I believe this is a side effect of #294.

When the app is first launched, it takes a while for the selection to be restored and the restoration is animated, which can confuse the user if the user scrolls before the restoration occurs.

Another issue is when switching tabs: if I have a saved selection for the schedule tab, it will be restored when coming back from the videos tab, destroying my scroll position. The restoration should only occur on app launch.

![scrollreset](https://user-images.githubusercontent.com/67184/30084624-8714ebf2-9269-11e7-9136-19164259424b.gif)

insidegui:
@VicenteBorrell Do you agree that the behavior shown in the gif is bad UX?

allenhumphreys:
Introduced in #294, it was definitely unintentional. I meant for it to scroll during an update to search results that are triggered by a user, though I was unsure whether that was even desired.

PR #309, opened 2017-09-06
Partially addresses issue #305

allenhumphreys:
This is the result of investigating #305. There are lot of moving parts during the app launch, and some duplicated/unnecessary calls that result in updating the search results multiple times.

1. `SearchCooridinator.configureFilters()` was/is getting called multiple times. The initial call is due the "offline" call, and the second call is in response to the synchronization callback. This may be related to #306, but I'm not sure.
2. `AppCooridinator.doUpdateLists()` was/is getting called multiple times. Same reason as 1

In this branch I addressed both of these defensively, which is less than ideal but may be required.

For case 1, in `SearchCooridinator`, I put checks in place to see if the "new" filter options are different than what is already configured on the controller. If they are the same, I do nothing. This was the quick approach as I can't be sure if there would be side effects if I enforced a "call once" policy on this method.

For case 2, in `SessionTableViewController`, I put checks in place to detect if the "new" array of sessions is identical to the old array (in content and in order), and if they are I simply don't do anything. The lack of this check was what was causing the weird scrolling after the user has scrolled. Basically, the list gets setup then the synchronization callback results in an update with identical values (accept in the case the the synchronization with the server might actually change some values).

Additionally, I noticed some major performance issues arising from the the code to update the table. I tracked most of it down to `SessionTableCellView` doing too much work synchronously in response to setting the `viewModel` property. Dispatching that asynchronously reduced noticeable UI delay when updating the filters considerably, it also **improves general scrolling performance 🎉**. You'll notice the thumbnails getting set **after** the row scrolls onto screen rather than before.

The other portion of delay was being caused by set operations, especially in the case when the old set and new set are both fairly large. I pushed everything onto the background to address that delay (0.1~0.4 seconds), but optimizing the looping logic might prove to be a better long term solution for that. As it stands, each loop is O(n^2) worst case since I call `index(of:)` for each element, and they might be the same elements.

### DIFF
--- a/WWDC/SessionTableCellView.swift
+++ b/WWDC/SessionTableCellView.swift
@@ -20,9 +20,11 @@ final class SessionTableCellView: NSTableCellView {
 
             disposeBag = DisposeBag()
 
-            thumbnailImageView.image = #imageLiteral(resourceName: "noimage")
+            DispatchQueue.main.async {
+                self.thumbnailImageView.image = #imageLiteral(resourceName: "noimage")
 
-            bindUI()
+                self.bindUI()
+            }
         }
     }
 
--- a/WWDC/FilterType.swift
+++ b/WWDC/FilterType.swift
@@ -46,9 +46,29 @@ extension Dictionary where Key == String, Value == [ String : Any ] {
                 //ToggleFilters
                 self[filterID.rawValue] = filter.dictionaryRepresentation()
             }
+        }
+    }
+}
 
+extension Array where Element == FilterType {
+
+    func isIdentical(to otherArray: [Element]) -> Bool {
+
+        var isIdentical = false
+
+        if self.count == otherArray.count {
+
+            isIdentical = true
+
+            for filter in self {
+
+                if !otherArray.contains(where: { $0.identifier == filter.identifier }) {
+                    isIdentical = false
+                    break
+                }
+            }
         }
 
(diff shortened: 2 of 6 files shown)


WWDC/SessionTableCellView.swift
```diff
@@ -20,9 +20,11 @@ final class SessionTableCellView: NSTableCellView {
 
             disposeBag = DisposeBag()
 
-            thumbnailImageView.image = #imageLiteral(resourceName: "noimage")
+            DispatchQueue.main.async {
+                self.thumbnailImageView.image = #imageLiteral(resourceName: "noimage")
 
-            bindUI()
+                self.bindUI()
```
allenhumphreys: This area was the primary bog down during scrolling and updates.

insidegui:
This looks amazing, I can't wait to try it out 👏

WWDC/SessionTableCellView.swift
```diff
@@ -45,8 +45,6 @@ final class SessionTableCellView: NSTableCellView {
 
         downloadedImageView.isHidden = true
         favoritedImageView.isHidden = true
-
-        thumbnailImageView.image = #imageLiteral(resourceName: "noimage")
```
allenhumphreys: Removing this seems to have fixed the no thumbnail issue

PR #310, opened 2017-09-08
Improve set display rows algorithm

otaviolima:
Improve #309 

I intentionally left the debug `prints` so we can see the difference. If PR commit get approved I will push a commit to this PR removing the `prints`.

@allenhumphreys Can you double check if that implementation has the same results as yours?


allenhumphreys:
@otaviolima This seems to do the trick! I had considered this general approach but couldn't ever force my brain to reason it out to completion!

No flickering, and it seems to me that it is still accurately reloading cells. And super snappy response times.🎉

insidegui:
Awesome! I modified it to suppress the logs in production builds. I'm preparing a release right now.

PR #312, opened 2017-09-12
Using icon in asset catalog from original images, potential fix for #311

### LINKED ISSUE #311: App icon shading rendered incorrectly on macOS High Sierra
screen shot 2017-09-09 at 10 48 02

The app icon renders in the dock & Finder with a lighter, "whitewashed" appearance on some devices running the macOS 10.13 beta. I think Gui & I initially assumed this was an OS issue that would be resolved, but seeing as it's persisted throughout betas & we're rapidly approaching the release of the OS its worth looking into whether we've made any colour profile or similar oversights.

My device is a 2016 MBP (13, 2). I contacted someone running 10.13 on an older MBP (11, 3), & they didn't appear to be having the same issue.

I found [similar reports for a handful other applications](https://github.com/atom/atom/issues/15410), but I struggled to recreate these now on my device, so they may have made an alteration without describing it in their issue tracking.

insidegui:
Potentially fixes #311 by adding icon to asset catalog from original PNGs


bcmn:
I'm away for work at the moment but I'll test this as soon as I get back. 😊

bcmn:
Great stuff.

PR #315, opened 2017-09-15
Increase contrast in the search field if the matching accessibility feature is turned on.

agatti:
If "Increase contrast" is turned on in the accessibility settings, macOS will use a darker shade of gray for text fields, assuming that the background is set to a light colour.  However, since WWDC uses a dark background, the text field becomes pretty much unreadable:

before_patch

After this patch the same control looks as such:

after_patch

The patch also makes the application listen for accessibility state changes and change the text colour accordingly.

Technically speaking, all of this may be sidestepped by just forcing the NSSearchField to always display using NSColor.white (or an appropriate light colour), but maybe that is not desirable.

### DIFF
--- a/WWDC.xcodeproj/project.pbxproj
+++ b/WWDC.xcodeproj/project.pbxproj
@@ -9,6 +9,7 @@
 /* Begin PBXBuildFile section */
 		012BEFA41EE8658F007E72CA /* EventKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 012BEFA31EE8658F007E72CA /* EventKit.framework */; };
 		01B3EB4A1EEDD23100DE1003 /* AppCoordinator+SessionTableViewContextMenuActions.swift in Sources */ = {isa = PBXBuildFile; fileRef = 01B3EB491EEDD23100DE1003 /* AppCoordinator+SessionTableViewContextMenuActions.swift */; };
+		6834F3771F6D227E004BB419 /* WWDCSearchField.swift in Sources */ = {isa = PBXBuildFile; fileRef = 6834F3761F6D227E004BB419 /* WWDCSearchField.swift */; };
 		DD0159A71ECFE26200F980F1 /* DeepLink.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0159A61ECFE26200F980F1 /* DeepLink.swift */; };
 		DD0159A91ED09F5D00F980F1 /* AppCoordinator+Bookmarks.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0159A81ED09F5D00F980F1 /* AppCoordinator+Bookmarks.swift */; };
 		DD0159CF1ED0CD3A00F980F1 /* PreferencesWindowController.swift in Sources */ = {isa = PBXBuildFile; fileRef = DD0159CE1ED0CD3A00F980F1 /* PreferencesWindowController.swift */; };
@@ -359,6 +360,7 @@
 /* Begin PBXFileReference section */
 		012BEFA31EE8658F007E72CA /* EventKit.framework */ = {isa = PBXFileReference; lastKnownFileType = wrapper.framework; name = EventKit.framework; path = System/Library/Frameworks/EventKit.framework; sourceTree = SDKROOT; };
 		01B3EB491EEDD23100DE1003 /* AppCoordinator+SessionTableViewContextMenuActions.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "AppCoordinator+SessionTableViewContextMenuActions.swift"; sourceTree = "<group>"; };
+		6834F3761F6D227E004BB419 /* WWDCSearchField.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = WWDCSearchField.swift; sourceTree = "<group>"; };
 		DD0159A61ECFE26200F980F1 /* DeepLink.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = DeepLink.swift; sourceTree = "<group>"; };
 		DD0159A81ED09F5D00F980F1 /* AppCoordinator+Bookmarks.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = "AppCoordinator+Bookmarks.swift"; sourceTree = "<group>"; };
 		DD0159CE1ED0CD3A00F980F1 /* PreferencesWindowController.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = PreferencesWindowController.swift; sourceTree = "<group>"; };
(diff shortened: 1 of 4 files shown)


bcmn:
This seems to correctly highlight user-input text, but I'm not sure the current destination for the code is the best place. If we're going to do a lot of operations on the search field cell maybe we should subclass NSSearchField & do it there instead of the view-controller.

Also, perhaps this could account for the magnifying glass-style search icon itself, which still goes to a very dark colour during high contrast, & the placeholder text when the window doesn't have focus, which is near unreadable with high contrast mode enabled. 

Otherwise though I'm really happy to see an accessibility-focused PR, it's an area I'd love to see more feedback & progress in for the app. If anyone visiting has any feedback in this area I think we'd all be really keen to hear it.

agatti:
Can you please tell me how to trigger the issue you're seeing?  I've spent quite some time to make sure it worked fine under all conditions - and from a cursory quick check the search field behaves correctly even when switching between applications and bringing the window out of focus.

I'll still subclass NSSearchField though as it does indeed make sense.  I'd love to work more on accessibility though and I'm glad there's a receptive audience for such a thing!

insidegui:
Thanks for the contribution @agatti ! I think @bcmn noticed the issue on High Sierra, maybe it's an OS thing?

Even though High Sierra is still in GM now, we're trying to support it because lots of people are already running it and many more will be updating later this month.

agatti:
That might be the case indeed.  I cannot really install a GM OS right now on this machine so I guess this PR should be left waiting until High Sierra is officially released so I would be able to reproduce the issue at hand.

However, if anybody with High Sierra is able to fix the problem in the meantime, I'm all for it :)

allenhumphreys:
Using macOS Sierra 10.12.6 my results differ from those in the original post.

Without increased contrast this branch:
screen shot 2017-09-16 at 8 42 00 pm

With increased contrast this branch:
screen shot 2017-09-16 at 8 42 09 pm

Without increased contrast master:
screen shot 2017-09-16 at 8 51 00 pm

With increased contrast master:
screen shot 2017-09-16 at 8 52 45 pm

agatti:
The plot thickens...  Can you please let me know what's your configuration for the "display" section of the accessibility settings?

Moreover, to get that result it means that the NSImage tinting function didn't work or was not called.  @allenhumphreys, can you please put a breakpoint on WWDCSearchField.tintedImage and make sure the right image comes out of that function?

agatti:
@allenhumphreys do you happen to have "Use dark menu bar and Dock" unchecked?

allenhumphreys:
Following up. I was testing with the assumption that the notification was functioning correctly, but it turns out `stopObservingAccessibilityOptionChanges` is being called due to a call to `viewWillMove(toWindow:)` with a value of `nil`. Though I'm not sure why that is happening.

When I enabled increase contract and then re-ran the app the image was tinted correctly.

Regarding "Use dark menu bar and Dock", I was attempting to test with and without it. It seems to have no effect on the final colors I'm seeing when testing on master.

agatti:
Ok, I seem to be able to replicate the issue by switching back and forth between "schedule" and "video" tabs.  Turns out there are two search fields rather than just one single view shared across tabs and viewWillMove(toWindow:nil) will be called to the one that is in the non-visible tab...

allenhumphreys:
I see, so maybe that was a red herring. My app isn't getting the notification no matter what I do. I commented out the `stopObservingAccessibilityOptionChanges` to make sure. I can't figure out why there is no notification 🤷‍♂️

agatti:
I can add a quick workaround to this - I'll just force the foreground colour to be white-ish at creation and then figure out why notifications are not sent around even though Apple's documentation explicitly says they should be.

allenhumphreys:
@agatti NSWorkspace has it's own notification center. You should update your code to use `NSWorkspace.shared().notificationCenter.addObserver...`

agatti:
Oh, I didn't know that.  Silly me...  Anyway, now I get notifications and whatnot, and all it's needed for it to work is to force a redraw once the image cell changes.  Stepping through Xcode works but if I trigger the changes myself with the Accessibility Inspector without having breakpoints the cell image is not repainted.  Apologies to everyone involved for all these inconveniences...

allenhumphreys:
No apologies necessary. This is teamwork! 😀

agatti:
Thanks for the review points, I've corrected them in my own version here (can you tell I'm quite new to Swift?), however for the life of me I cannot seem to be able to force an image update of the NSSearchField image cell if the window is not key.  The cell does get repainted correctly when the window is key again.  Right now what I have is this (I've been adding methods as I tried):

```swift
        self.needsDisplay = true
        self.setNeedsDisplay()
        self.window?.contentView?.needsDisplay = true
        self.window?.contentView?.setNeedsDisplay(self.frame)
        self.window?.displayIfNeeded()
```

but nothing seems to work - however the text colour does change when the window is not key so there's something I've been missing but I have no idea what it might be.

allenhumphreys:
Yeah, it looks like you've got a real tricky issue. The image gets updated as soon as the window regains focus. I don't think you're missing anything, I just think you may've found a quirk of NSSearchField. It seems to be failing to redraw its `searchButtonCell` appropriately. Actually, if you remove your image adjusting code entirely and just observe what the standard system behavior is, you'll see that the image doesn't change until the window receives focus. I think this is the strongest evidence to suggest its a system bug. All-in-all, the behavior isn't terrible in that the image does update as soon as the user re-focuses the window. 

The best approach might be to simply make the magnifying glass the white color, or simply a lighter color all in all situations instead of trying to change it dynamically. You can see my screen shot from master, the default color is already pretty light. Measuring it, it is effectively (222, 222, 222), so you should make sure to keep that behavior in the default state at the very least, currently you're setting it to the `controlTextColor` which is resulting in the magnifying glass icon being black for me.

@bcmn Thoughts?

agatti:
Forcing both the cell and the text to always have a lighter colour at all times is another option indeed.  From a few quick tests, when "increase contrast" is on, intensity is boosted by 20% or so - tinting the search cell to #B7B7B7 and setting the text to #BEBEBE will, once contrast is increased, make text look something around #C7C7C7.

I found nothing on openradar regarding NSSearchField, however.  If the same behaviour is observed on High Sierra too then I'll go ahead and file one myself.

bcmn:
I think the behaviours described here are consistent with what I'm seeing. I was having issues with build times when first reviewing this so it was quite hard to compare behaviours well, especially as a fairly complex mix seems to be involved here. 

> Actually, if you remove your image adjusting code entirely and just observe what the standard system behavior is, you'll see that the image doesn't change until the window receives focus. I think this is the strongest evidence to suggest its a system bug.
> I found nothing on openradar regarding NSSearchField for similar issues, however. If the same behaviour is observed on High Sierra too then I'll go ahead and file one myself.

This definitely happens on HS.

bcmn:
> The best approach might be to simply make the magnifying glass the white color, or simply a lighter color all in all situations instead of trying to change it dynamically.

I think this would be the best/simplest option at this point? I don't think it's especially erroneous for the icon to look _slightly_ different to the text, & to me consistency would be more important. It's a lot nicer than the current change depending whether increased contrast is enabled before or after the app is started, anyway.

This system-enforced change is rather irritating, as without it the contrast does meet WCAG 2.0 guidelines by quite a way. 🤔

insidegui:
@agatti Have you been able to check this on High Sierra?

Issue #296, opened 2017-08-30
How one can run the tests?


otaviolima:
I was trying to run the tests but encountered a lot of compiler errors and after fixing some and commenting out others I still was not able to run.

From the error is pretty obvious that is due some code signing process.

My questions then are:

Would be a person not enrolled as mac developer in Apple Developer Center able to run the tests?

If so, is the current code correct and compiling? At least for me, as I mentioned earlier, it did not compile.

insidegui:
Unfortunately, the tests are currently very outdated and are not compiling. I developed ConfCore (data parsing and storage) a long time ago with tests and when the time came to actually put the app into production I didn't have enough time to do proper TDD or update the tests. To make that even worse, a few days before WWDC Apple changed the data structure completely and that completely invalidated the tests (and made I leave some messy code in there because of the time pressure).

SpacyRicochet:
Sounds like a nice [Hacktoberfest[(https://hacktoberfest.digitalocean.com) issue. Consider adding that label to some of the issues here.

Ethenyl:
Hi there,

I'm more than willing to help on that topic, if it's OK on your side I'll create a PR as soon as I have something stable.

insidegui:
@Ethenyl That would be great, go ahead 🚀

Issue #318, opened 2017-09-24
setDisplayedRows calls UI method from a background queue


insidegui:
The `setDisplayRows` method in `SessionsTableViewController.swift` calls `tableView.selectedRowIndexes` from a background queue. Although this might not be causing any issues currently, it could result in weird behavior and/or random crashes.

https://github.com/insidegui/WWDC/blob/d69676e63924ce47195850bde821d8471c97a697/WWDC/SessionsTableViewController.swift#L83

antonselyanin:
I moved this call to the main thread in #322 PR because unit tests were crashing.

insidegui:
#322 disabled main thread checking so the tests would run. When this issue gets fixed, we should re-enable the main thread checker for tests.

PR #322, opened 2017-10-15
Fixed unit tests compilation and running

antonselyanin:
Fixed tests launching for the issue #296

* Disabled code signing for the test target
* Renamed `DatabaseTests` to `StorageTests`, I think it makes a bit more sense.
* Didn't touch `NetworkingTests`, I figured it requires a local server. IMO, it's better not to rely on such dependencies in unit tests. We could move them to some sort of integration tests.

Implementation question:

In `Storage.makeRealm` a realm configuration was being recreated from `realmConfig`

```
        let c = Realm.Configuration(fileURL: realmConfig.fileURL, schemaVersion: realmConfig.schemaVersion, migrationBlock: realmConfig.migrationBlock)

        return try Realm(configuration: c)
```
I couldn't come up with a reason why it was necessary. Did I miss something?
I had to remove this logic and use `realmConfig` instead, so that I could use in-memory realm.

Further improvements:
- write more tests for ConfCore
- set up CI (otherwise tests will inevitably rot)

How do you think, do tests have a future in the project? :D


insidegui:
Thanks for the PR @antonselyanin 🎉

@bcmn is going to review the changes for us.

> Didn't touch NetworkingTests, I figured it requires a local server. IMO, it's better not to rely on such dependencies in unit tests. We could move them to some sort of integration tests.

Should we even have that? 🤔 The networking is all handled by Siesta and the things that can go wrong are related to parsing/storing (which we are testing).

> How do you think, do tests have a future in the project? :D

Definitely, especially unit tests. I love being able to change model stuff and run a test to be sure it all still works 😊

WWDC/SessionsTableViewController.swift
```diff
@@ -80,7 +83,7 @@ class SessionsTableViewController: NSViewController {
                 return
             }
 
-            let selectedRows = self.tableView.selectedRowIndexes.flatMap { (i) -> IndexedSessionRow? in
+            let selectedRows = selectedRowIndexes.flatMap { (i) -> IndexedSessionRow? in
```
allenhumphreys: Moving this outside the lock is technically wrong. This code block can be move down inside the `sync` block, however, maintaining data integrity and calling the API on the main thread.

insidegui: Thanks for pointing that out. @antonselyanin can you change it following @allenhumphreys suggestion?

allenhumphreys: Just a small follow up, it’s not just this code block, there are a couple pieces of logic that are only related to preserving selection that are scattered through the function. They can all be moved together. The only reason they’re separate was to make the code look pretty 🤓. I’d offer more specific guidance but I’m stuck with just my phone this weekend!

antonselyanin: I found another solution - I'll just disable Main Thread Checker for 'Test' mode and revert `SessionsTableViewController` changes.
'Main Thread Checked' will still be on for normal app runs.

It would be better to address `selectedRowIndexes` issue separately.

bcmn:
This seems good now, thanks all. 👍 We've added a note to #318 to revert the Main Thread Checker change as soon as that issue is solved.

PR #323, opened 2017-10-15
Fix async UI access

### LINKED ISSUE #318: setDisplayedRows calls UI method from a background queue

bcmn:
Fixes #318.
- [x] Moved selected row access to main thread.
- [x] Re-enabled thread checker for debug.

I paused a couple times adding some of this code to the main thread, but in practice the time difference seems to be negligible.


insidegui:
@allenhumphreys Can you check this?

Thanks

PR #334, opened 2018-01-16
Remove IGListKit dependency

otaviolima:


otaviolima:
Don't merge yet. It is missing one file.

PR #336, opened 2018-01-22
Adopt SwiftLint and CI

bcmn:
I think @insidegui & I have been keen on SwiftLint for a while, & it definitely comes up occasionally with contributors who are used to projects with it. As such, I thought it might be useful to adopt it in such a way as to try & cause minimum work for our current codebase, & we can make it more strict as time goes on & we have more time to refactor.

Currently disabled rules:
- line_length
- function_body_length
- file_length
- type_body_length
- identifier_name
- function_parameter_count
- todo
- fallthrough
- unused_closure_parameter

The majority of the current selection are there not as a comment on our desired style, but to try & avoid this initial PR being a more significant refactor, instead of just a basic tidy-up. Obviously these could be addressed & narrowed down a lot further before merging. The final three were disabled as it seemed as though the instances they found could be considered pretty valid, explicit choices of the original programmer, or which make the codebase more explicit about what's happening. 🤷🏻‍♂️ Discuss!

Feel free to compare & contrast with [SwiftLint's full collection of rules](https://github.com/realm/SwiftLint/blob/master/Rules.md), & suggest any to explicitly enable or disable -- I can't claim to have comprehensively gone through every item.


allenhumphreys:
I, for one, think this should be pulled in sooner rather than later! The improvements to style enforcement are great! My only thought is that without CI in place to fail PRs (unless I missed something?), the script phase should fail if SwiftLint isn't installed. Additionally, we should warn if the user's SwiftLint version is behind the version the project is expecting.

insidegui:
I agree

bcmn:
Okay, build now fails if not installed, & warns if less than the minimum version (which I simply based on the version it appears I started this branch with).

I also quickly added to `bootstrap.sh` to handle all this when the project is initially cloned & initialised.

bcmn:
Just to summarise a discussion here, Gui & I were going over it & we still have a few changes to make to `bootstrap.sh` before this is merged.

bcmn:
Hm. ☝🏼 This commit only seems to have failed because the Storage Tests either hung or stopped outputting logs 🤔

bcmn:
Hm, it's succeeded now, but I'm a little nervous about whether this will be a semi-regular occurrence if we merge 😕

Otherwise, what would we need to do to get this PR merged, @allenhumphreys? I feel like any improvements to the CI implementation could easily be other PRs.

allenhumphreys:
@bcmn It's already proved itself to be a semi regular occurrence. It failed in that manner for me twice before, I'll spend today trying to debug it and report back.

allenhumphreys:
@bcmn I actually managed to get the test to hang locally (it's a miracle). Here is what seems to be causing it:
screen shot 2018-03-08 at 9 43 54 am

Pressing OK let the test proceed. You can reset the Sparkle update interval by `defaults delete io.wwdc.app SULastCheckTime` and then run the test. 

I've disabled update checking in debug builds since they always seem to result in this error anyway. I could disable updates just for tests but this method seems simpler.

allenhumphreys:
Additionally...

> I feel like any improvements to the CI implementation could easily be other PRs, so long as the current implementation is reliable.

I agree 👍

insidegui:
> I've disabled update checking in debug builds since they always seem to result in this error anyway. I could disable updates just for tests but this method seems simpler.

That's be best approach, thanks for doing that!

> I feel like any improvements to the CI implementation could easily be other PRs, so long as the current implementation is reliable.

I agree.

Additionally, I feel like after this gets merged it'll be easier for us to figure out which rules to keep and which rules to change and get contributions fixing some of the warnings 😅

@bcmn Wanna do the honors?

bcmn:
Awesome stuff, thanks very much guys! 🎉

PR #337, opened 2018-01-22
Adopt Code of Conduct

bcmn:
Adopts the [Contributor Covenant](https://www.contributor-covenant.org) code of conduct. We want this project to be as approachable as possible for all, & we should be making that as clear as we can.


insidegui:
Thanks!

PR #342, opened 2018-03-01
Update dependencies and address API changes

allenhumphreys:
Keeping these up to date will keep build times reasonable. I also locked in a specific version for protocol buffers because of the ChromeCastCore dependency.


WWDC/PlaybackViewModel.swift
```diff
@@ -34,7 +34,8 @@ extension Session {
         if type == .image {
             return imageAsset()
         } else {
-            return assets.filter({ $0.assetType == type }).first
+            let filtered = assets.filter({ $0.assetType == type })
```
allenhumphreys: I'm not sure why the compiler needs this separated. If they're chained it gives an ambiguous function reference error.

insidegui: This happens, sometimes I can't infer the types correctly ¯\_(ツ)_/¯

PR #344, opened 2018-03-08
Address all warnings except cyclomatic complexity

allenhumphreys:
I added another `.swiftlint.yml` to the `PlayerUI` directory which disables 2 rules: cyclomatic_complexity, block_based_kvo. I did this because the player part of the app is quite complicated and it could take a considerable amount of time to safely refactor the affected parts.

I addressed the outstanding warnings for discarded Notification Center observers through a combination of explicitly ignoring the token in the case of singletons, using RxSwift to manage the lifetime, and using the object based tokens to manage the life time. I was unsure whether we wanted some classes to have a dependency on RxSwift just to manage notification lifetimes.

I address the KVO warnings, but I'm not sure I really like the result, thoughts?

I disabled the force_cast and force_try rules for certain issues. A different option would be to create a `castOrFatalError` helper function similar to the `tryOrXCTFail` function I created for silencing the warnings in the tests.

### DIFF
--- a/WWDC/AppCoordinator.swift
+++ b/WWDC/AppCoordinator.swift
@@ -91,9 +91,9 @@ final class AppCoordinator {
         setupBindings()
         setupDelegation()
 
-        NotificationCenter.default.rx.notification(.WWDCTabViewControllerDidFinishLoading).subscribe(onNext: { [weak self] _ in self?.restoreApplicationState() }).disposed(by: disposeBag)
-        NotificationCenter.default.rx.notification(NSApplication.didFinishLaunchingNotification).subscribe(onNext: { [weak self] _ in self?.startup() }).disposed(by: disposeBag)
-        NotificationCenter.default.rx.notification(NSApplication.willTerminateNotification).subscribe(onNext: { [weak self] _ in self?.saveApplicationState() }).disposed(by: disposeBag)
+        _ = NotificationCenter.default.addObserver(forName: .WWDCTabViewControllerDidFinishLoading, object: nil, queue: nil) { _ in self.restoreApplicationState() }
+        _ = NotificationCenter.default.addObserver(forName: NSApplication.didFinishLaunchingNotification, object: nil, queue: nil) { _ in self.startup() }
+        _ = NotificationCenter.default.addObserver(forName: NSApplication.willTerminateNotification, object: nil, queue: nil) { _ in self.saveApplicationState() }
     }
 
     /// The list controller for the active tab
@@ -233,7 +233,7 @@ final class AppCoordinator {
         resetAutorefreshTimer()
     }
 
-    func startup() {
+    private func startup() {
         RemoteEnvironment.shared.start()
 
         ContributorsFetcher.shared.load()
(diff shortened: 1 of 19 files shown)


WWDC/AppCoordinator.swift
```diff
@@ -91,9 +91,9 @@ final class AppCoordinator {
         setupBindings()
         setupDelegation()
 
-        NotificationCenter.default.addObserver(forName: .WWDCTabViewControllerDidFinishLoading, object: nil, queue: nil) { _ in self.restoreApplicationState() }
-        NotificationCenter.default.addObserver(forName: NSApplication.didFinishLaunchingNotification, object: nil, queue: nil) { _ in self.startup() }
-        NotificationCenter.default.addObserver(forName: NSApplication.willTerminateNotification, object: nil, queue: nil) { _ in self.saveApplicationState() }
+        NotificationCenter.default.rx.notification(.WWDCTabViewControllerDidFinishLoading).subscribe(onNext: { [weak self] _ in self?.restoreApplicationState() }).disposed(by: disposeBag)
```
insidegui: Given that the AppCoordinator is essentially a singleton (it lasts for the lifetime of the app and there's only ever one), I think we can safely ignore the notification observer token.

WWDC/VideoPlayerViewController.swift
```diff
@@ -177,15 +173,22 @@ final class VideoPlayerViewController: NSViewController {
...
+    private var presentationSizeObserver: NSKeyValueObservation?
+    private var currentItemObserver: NSKeyValueObservation?
 
-        if keyPath == #keyPath(AVPlayer.currentItem.presentationSize) {
-            DispatchQueue.main.async(execute: playerItemPresentationSizeDidChange)
-        } else if keyPath == #keyPath(AVPlayer.status) {
-            DispatchQueue.main.async(execute: playerStatusDidChange)
-        } else {
-            super.observeValue(forKeyPath: keyPath, of: object, change: change, context: context)
+    private func setupPlayerObservers() {
```
insidegui: Have you checked if the usage of the new KVO API is fine here? I remember trying to use it before and giving up because AVFoundation objects did not work well with it.

WWDC/VideoPlayerViewController.swift
```diff
@@ -177,15 +173,22 @@ final class VideoPlayerViewController: NSViewController {
...
+
+        playerStatusObserver = player.observe(\AVPlayer.status, options: [.initial, .new], changeHandler: { [weak self] (player, change) in
+            guard let strongSelf = self else { return }
```
insidegui: I've been using `guard let 'self' = self` (pretend the ' are backticks) in these cases lately, maybe we should standardize and use it throughout?

insidegui:
Wow, that was FAST! Haha!

Thank you very much for addressing this. Have a look at the comments and let me know what you think.

WWDC/VideoPlayerViewController.swift
allenhumphreys: I think I love this! Never considered it. Way less prone to people coming along and adding `self?.whatever` alongside the `strongSelf.otherthing` and getting the context all confused.

WWDC/VideoPlayerViewController.swift
allenhumphreys: I went back to gather some info on this question and found that the function calls were reversed, status change is calling `playerItemPresentationSizeDidChange` and vice versa. Whoops!

The information I gathered is simply this:

```
new status did change
new presentation size did change
new status did change
old status did change
old presentation size did change
old status did change

new presentation size did change
old presentation size did change

new presentation size did change
new presentation size did change
old presentation size did change
old presentation size did change

new presentation size did change
new presentation size did change
new presentation size did change
old presentation size did change
old presentation size did change
old presentation size did change

new presentation size did change
old presentation size did change

new presentation size did change
old presentation size did change
```

All the calls are balanced, though the order of the calls makes me 🤔

WWDC/VideoPlayerViewController.swift
insidegui: Yeah and it leaves the syntax highlighting intact as well. Win/win/win 😄

allenhumphreys:
@insidegui These changes were relatively minor, especially because I just ignored some of them 🤫. 

Addressing the cyclomatic complexity warnings might take a bit longer.... At least one of them I think should be ignored because it seems to be counting guards inside of blocks as part of the execution paths of a function, see `downloadStatusObservable`.

PR #354, opened 2018-03-14
Fixes `viewDidAppear` not being called on SessionsTableViewController

allenhumphreys:
During my investigation into applying restored filters during the initial load of data rather than after, I found this problem.

1. Quit the app with the videos tab selected so that it is now the saved tab
2. Relaunch the app

At this point, `viewDidAppear` will be called on both tabs even though the schedule tab isn't selected. Additionally, `viewDidAppear` will never be called again for the schedule tab.

I tracked the root of problem down to setting the active tab within the execution context of `viewDidAppear` on `WWDCTabViewController` which was being caused by the firing of the **WWDCTabViewControllerDidFinishLoading** notification. Dispatching the notification asynchronously fixed the problem, but I was curious why the notification mechanism existed in the first place.

It seems it was in use to facilitate centering the toolbar items, so I did some research and found that as of 10.10 the preferred way of doing stuff like this is to override [`toolbarDefaultItemIdentifiers(:)`](https://developer.apple.com/documentation/appkit/nstabviewcontroller/1428251-toolbardefaultitemidentifiers).

> For example, include flexibleSpace strings as the first and last elements of the array to center the remaining toolbar items.

Once I found that, I was able to remove the notification altogether!

### DIFF
--- a/WWDC/SessionsTableViewController.swift
+++ b/WWDC/SessionsTableViewController.swift
@@ -193,8 +193,6 @@ class SessionsTableViewController: NSViewController {
         selectSession(with: identifier, scrollOnly: true)
     }
 
-    private var setupDone = false
-
     override func viewDidAppear() {
         super.viewDidAppear()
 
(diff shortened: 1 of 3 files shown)


WWDC/SessionsTableViewController.swift
```diff
@@ -205,15 +203,21 @@ class SessionsTableViewController: NSViewController {
...
+    /// This function is meant to ensure the table view gets populated
+    /// even if its data model gets added while it is offscreen. Specifically,
+    /// when this table view is not the initial active tab.
     private func performFirstUpdateIfNeeded() {
+        guard !setupDone else { return }
+        setupDone = true
+
         switch style {
         case .schedule:
             updateScheduleList()
```
allenhumphreys: Notice the missing guard here. This bug was being hidden by the appearance method bug because generally outside of WWDC week, the schedule tab is not the saved tab. But if the schedule tab is the saved tab, this bug presented itself by reloading the table every time you would visit the schedule tab.

insidegui:
Nice catch! Does this fix #355 ?

allenhumphreys:
@insidegui I don't think so, but I have something else cooking to address that. I wanted to merge this separately though as this other thing is a bit riskier!

insidegui:
Cool :)

PR #356, opened 2018-03-15
Switch from Timer to NSBackgroundActivityScheduler

allenhumphreys:
There are some quirks as compared to using a Timer, specifically the fact that the activity may or may not run immediately when scheduled.

I also hooked up the preference so that the timer gets created immediately rather than needing to relaunch the app.

### DIFF
--- a/WWDC/LiveObserver.swift
+++ b/WWDC/LiveObserver.swift
@@ -16,7 +16,14 @@ final class LiveObserver {
     private let dateProvider: DateProvider
     private let storage: Storage
 
-    private var timer: Timer?
+    private lazy var backgroundActivity: NSBackgroundActivityScheduler = {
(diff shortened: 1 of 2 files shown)


insidegui:
Great refactor 🎉

WWDC/LiveObserver.swift
```diff
@@ -63,7 +67,7 @@ final class LiveObserver {
         return storage.realm.objects(SessionInstance.self).filter("isCurrentlyLive == true")
     }
 
-    @objc private func checkForLiveSessions(_ sender: Any?) {
+    private func checkForLiveSessions() {
         let startTime = dateProvider()
```
otaviolima: Love seeing those `@objc` disappearing from code base :D

PR #357, opened 2018-03-16
Apply restored search settings during intialization and other fixes

### LINKED ISSUE #355: Selection restoration is broken.

- Version of macOS: 10.13.3
- Version of the app: 5.3
- If it is a wrong behavior, list all steps required to reproduce it

1. Launch the app
2. Select a video not at the top of the list
3. Quit the app via Cmd + Q to ensure state is saved
4. Relaunch the app

Expected Result:
The selected video upon exit is reselected on app launch

Actual Result:
The top video in the list is selected.

allenhumphreys:
* Fixes #355 - selection restoration
* Applies the saved filters during app startup's initial displaying of data, avoiding the ugly animation
* Fixes the flickering behavior of the search fields by reworking how we prevent them from automatically becoming first responder (previously we were setting the first responder to nil when the `SessionsTableViewController` appeared sometimes resulting in seeing parts of the search fields animations)
* Simplifies list restoration by moving deferral into `SessionsTableViewController` so it can handle its own initialization
* Hides all UI initialization, animating the `alphaValue` up when the list has been initialized. No scrolling should be visible, though some thumbnails may appear to load after the table is visible
* Simplifies `SessionsTableViewController` by creating `SessionRowProvider` to abstract away the fact that SessionsTableViewController was actually doing 2 different jobs
* Fixes app crashes during WWDC week caused by #356 
* When performing automatic scrolling during search and state restoration, the row will now be centered in the table if possible

Things I Tested:
* Behavior of fresh install by using `cleardata.sh`
* Deeplink handling both when the app is running and when the app gets launched
* During WWDC using --delorean

### DIFF
--- a/WWDC/AppCoordinator.swift
+++ b/WWDC/AppCoordinator.swift
@@ -284,8 +284,12 @@ final class AppCoordinator {
 
     private func saveApplicationState() {
         Preferences.shared.activeTab = activeTab
-        Preferences.shared.selectedScheduleItemIdentifier = selectedScheduleItemValue?.identifier
-        Preferences.shared.selectedVideoItemIdentifier = selectedSessionValue?.identifier
+        if let identifier = selectedScheduleItemValue?.identifier {
+            Preferences.shared.selectedScheduleItemIdentifier = identifier
+        }
+        if let identifier = selectedSessionValue?.identifier {
+            Preferences.shared.selectedVideoItemIdentifier = identifier
+        }
         Preferences.shared.filtersState = searchCoordinator.currentFiltersState()
     }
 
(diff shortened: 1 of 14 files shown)


WWDC/AppCoordinator.swift
```diff
@@ -280,59 +282,48 @@ final class AppCoordinator {
...
     // MARK: - Deep linking
 
-    func handle(link: DeepLink) {
-        guard didRestoreLists else {
-            deferredLink = link
-            return
-        }
+    func handle(link: DeepLink, deferIfNeeded: Bool) {
 
         if link.isForCurrentYear {
```
insidegui: Nice! I like this behavior

PR #361, opened 2018-03-19
Fixes incorrect behavior of session context menu

### LINKED ISSUE #347: Context menu doesn't handle watched/unwatched correctly

- Version of macOS: 10.13.3
- Version of the app: 5.3
- If it is a wrong behavior, list all steps required to reproduce it
When you watch a video, the context menu will never treat the video as "watched" even if you watch it to the absolute end.

Expectation: Videos filtered out by the "unwatched" filter should also show "Mark as Unwatched" in the context menu.

note: The "Mark as Unwatched" will be activated if you first "Mark as Watched" which will set the relative progress to exactly 1. While the playback time observer sets the progress to things like 0.998868563034773 and 1.00000097433287.

I believe the context menu should use the same threshold as the "unwatched" filter, which is 0.9. This could be achieved by adding an `isWatched` non-stored property to `Session` to encapsulate the handling. It also looks like the relative position could use a little extra math to cap it at 1.0 and possibly do a little rounding?

allenhumphreys:
Fixes #347 

It turns out there was a constant already defined, it's value is 0.97.

### DIFF
--- a/WWDC/SessionsTableViewController.swift
+++ b/WWDC/SessionsTableViewController.swift
@@ -471,9 +471,13 @@ class SessionsTableViewController: NSViewController {
 
         switch menuItem.option {
         case .watched:
-            return viewModel.session.progresses.first == nil || viewModel.session.progresses.first?.relativePosition != 1
+            let canMarkAsWatched = !viewModel.session.isWatched
+                && viewModel.session.instances.first?.isCurrentlyLive != true
+                && viewModel.session.asset(of: .streamingVideo) != nil
+
+            return canMarkAsWatched
         case .unwatched:
-            return viewModel.session.progresses.first?.relativePosition == 1
+            return viewModel.session.isWatched
         case .favorite:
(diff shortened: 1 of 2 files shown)


WWDC/SessionsTableViewController.swift
```diff
@@ -497,6 +501,17 @@ class SessionsTableViewController: NSViewController {
     }
 }
 
+extension Session {
+
+    var isWatched: Bool {
```
allenhumphreys: In your opinion, where should this live @insidegui ? The `Constants` are part of WWDC, while `Session` is part of ConfCore. So it feels fine as an extension defined in WWDC but could probably use a better home.

insidegui: Interesting question. I think we can leave it as is by now, but maybe some constants should be moved into ConfCore (or have some entity in ConfCore that can hold these types of configurations).

PR #363, opened 2018-03-24
Narrows the scope of the ATS disabling

allenhumphreys:
As far as I could tell, the only exceptions for ATS are the test environment "http://localhost:9042" which is covered by `NSAllowsLocalNetworking` and `asciiwwdc.com` which is given an explicit exception via `NSExceptionDomains`.


insidegui:
Maybe we should have an exception for media as well, not all video content (especially live) is served over https.

allenhumphreys:
@insidegui Good point! Updated!

PR #368, opened 2018-04-28
Initial Touch Bar implementation for the video player

insidegui:
This is the initial Touch Bar implementation for video playback controls. The Touch Bar includes all basic player commands and is fully customizable. Deployment target has been raised to 10.12.2 to avoid having to check for API availability all the time.

touch bar shot 2018-04-28 at 15 46 07


VicenteBorrell:
Most of the icons look blue-ish, is that normal?

allenhumphreys:
@vincenteborrell There definitely seems to be different colors! The skip back/forward, and play buttons seem to be a little redder. The next/previous buttons seem to more closely match the escape button

VicenteBorrell:
Maybe it has something to do with the color space of the icon files.

bcmn:
☝🏻 It's very subtle, but I can corroborate the colour issue above.

Functionally it's really good though, handles things like switching the skip time really well. Could we change the PiP icon to the "Exit PiP" icon once in PiP mode? Similarly, this makes me wonder if there could be some additional feedback when adding a bookmark. I'm not sure of an ideal solution here, though -- fading up the playback bar briefly might be frustrating as a user? 

Also, bluetooth media button support! 😍🎉

VicenteBorrell:
Sure! BTW I wasn't trying to nitpick or anything ;)

insidegui:
Ok the color issue is now fixed 🎉

I've enabled template mode for the Touch Bar icons, which is the recommended approach since the Touch Bar can then adapt the color of the icons to the ambient light color. You can see the difference on the images below:

Cold lighting:
touch bar shot 2018-05-06 at 10 47 14

Warm lighting:
touch bar shot 2018-05-06 at 10 49 33

@VicenteBorrell How are we doing on those icons? (exit pip and exit fullscreen)

WWDC/VideoPlayerViewController.swift
```diff
@@ -122,6 +122,12 @@ final class VideoPlayerViewController: NSViewController {
         }.disposed(by: disposeBag)
     }
 
+    override func viewDidAppear() {
+        super.viewDidAppear()
+
+        view.window?.makeFirstResponder(playerView)
```
allenhumphreys: This conflicts with `SessionsTableViewController.viewDidAppear` such that changing tabs doesn't result in player becoming first responder. Actually, the touch bar controls appear then disappear.

I think the table view wants first responder so you can move around the list with the keyboard... possibly? But, the explicit `makeFirstResponder` call can be removed from `SessionsTableViewController` because in `AppCooridinator` I did this:

```
scheduleItem.initialFirstResponder = scheduleController.listViewController.tableView
```

Which I believe made the call unnecessary and allows you to keep this hear.

insidegui: Really? On my tests, the player view successfully became the first responder, activating the Touch Bar. It has the side effect of stopping arrow navigation on the sidebar when you select the currently playing item tho, I'm not sure which one is worse (breaking arrow navigation or not having the Touch Bar controls show up automatically) :|

allenhumphreys: Hey, just got the chance to double check. Switching tabs definitely causes the controls to flash on then back off as the table view takes first responder back.

Regardless of that, I like where you're heading. I agree that it makes sense that the controls be visible anytime a video is playing unless there are plans to have other touch bar controls for different app interactions.

insidegui: > unless there are plans to have other touch bar controls for different app interactions

Touch Bars can be combined, so that's not an issue :)

VicenteBorrell:
I just uploaded the PiP Exit icon to the UI Kit artboard on Zeplin. Let me know if you need anything else 😉 
I was thinking about redoing the player icons and fix some small details. Maybe next weekend!

insidegui:
Awesome! Can you make an "exit full screen" button as well? We can replace the Touch Bar's escape key with it when in full screen 😊

insidegui:
@allenhumphreys My latest push has the change to the responder chain stuff we talked about, test it if you can and tell me what you think 😉

Issue #331, opened 2017-12-22
Feature Request: Display a List of Related Videos


nik6018:
Hi gui,

Thanks for the amazing Mac App the official iOS version is horrible, I have been using the App for a week and think the following may make the UX better.

It would be great if the app can have a related video's section under the Main Video view.

The WWDC app for iOS does this and this is very helpful for finding related videos on a topic that you watch.

Otherwise, I have to search for the Video first on the iOS app check the related videos and then search for the exact name of the session in the Mac App.

Don't know if this can be implemented if you are just scrapping Apple's website for videos, hoping for the best.

Would love to help to get this feature implemented if it's possible.

insidegui:
This data is actually available from the API, so it can definitely be done 😄

nik6018:
Thanks gui.
Hope an amateur like me can help in some way.

bcmn:
Worth noting here that I've begun working on the JSON parsing for this, & we've requested an update from the UI designer to take it into account. I was on a train so I couldn't push the changes, but will ASAP.

insidegui:
Thanks @bcmn !

VicenteBorrell:
Hi guys! It's been a while! 😄 

I just made this two proposals for this feature. Let me know what you think.

![related videos section](https://user-images.githubusercontent.com/584834/37118023-239a6776-2253-11e8-84c7-fe6112b0a230.png)
![overview with related](https://user-images.githubusercontent.com/584834/37118025-23c313e2-2253-11e8-8b89-3676a8ae4d27.png)

allenhumphreys:
I like the second option. Is it part of the overview "tab" or will it always be visible?

Also, I wanted to mention that if we're change this area of the UI, you might want to consider issue #233. @insidegui Is there an easy way to make the text scrollable and also automatically resize like it does today?

bcmn:
I’ve looked into #233 before and not made much progress, but I’m not very
au fait with AppKit (& I’m not sure an already implemented app with is
necessarily the best way to learn 🤷🏻‍♂️). There’s some real weirdness
with how the video & the text section below scale in relation to each other.

For what it’s worth, I’d agree whole-heartedly with the second option — I’m
not even slightly convinced people would take the time to click through to
another tab. I think my instinctive expectation from looking at the mock-up
would be that if there was a really long summary, it would be scrollable &
fade “behind” the related videos, with the core details (session no.,
focuses, etc.) at the bottom of the scrollable summary (as in, if it was
[...]

Issue #329, opened 2017-10-17
Old database versions for migration tests


antonselyanin:
I could write some unit tests for `Storage` class, I wanted to start with the migration logic.
Do you have realm files with old database schemas? 

In case you don't have old databases. How do you think, would "reversing" the schemas from the migration code be sufficient for unit tests?

bcmn:
Migration tests seem like a good idea --even if we're a little late to the game on this one.

I don't know if others have old databases they use, but its worth noting that the last V4 release is still available [here](https://github.com/insidegui/WWDC/releases/tag/4.3.1). There's no need for reverse engineering the structure from the code we're going to be testing, and doing so seems like it'd only serve to defeat the purposes of the unit test by perpetuating any wrong assumptions. 🤔

insidegui:
Well, we do have some cases (#284 ) where the migration seems to be causing problems. If we could isolate that with unit tests maybe it could help with the solution. @antonselyanin Please send me your e-mail so I can invite you to our Slack.

insidegui:
Do we still think this is important/useful?

It's been almost a year since the 5.0 update, I feel like we should plan on removing the migration code altogether at some point.

bcmn:
Yeah, I think it’d be perfectly reasonable to remove the migration code with 6.0. Though I feel we should still try to make it a principle of the project to test any (presumably smaller) schema migrations in future.

ChinaHackers:
The macOS 10.12.6 WWDC5.3 version prompt does not respond.
Why does it happen?

insidegui:
@ChinaHackers Please [open a new issue](https://github.com/insidegui/WWDC/issues/new). Thanks!

ChinaHackers:
OK!

ChinaHackers:
@insidegui OK.

PR #367, opened 2018-04-26
Parse and persist related videos and resources

bcmn:
Parses and persists the data relating to related resources (such as sample code, documentation, guides or other sessions) for each session. This lays the groundwork for the UI side of #331.

When a Session is parsed it gets a collection of placeholder SessionResource objects, each of which is resolved to a RelatedResource object in Storage.swift, which contains all the information needed for the UI. The UI is anticipated to access this information by iterating over `Session.related`.

---

I had been planning a slightly more significant refactor of this, primarily to remove the new SessionResource type completely, but it'd been on & off & Gui's new tests seem to have shown I've missed some behaviour in my progress on that so I stashed and I'll rethink that soon.

PR #369, opened 2018-04-29
Not messing with session identifiers, refactored database migration code

insidegui:
As pointed out in #331 by @bcmn, the app was replacing `"wwdc"` in session identifiers for some unknown reason and this could cause issues with different events (like the `"fall"` event).

With this PR, I removed that unnecessary replacement, we're now using the identifiers as they come from the API. I also moved the `year` property from `Session` to `Event` as a stored property so it can be queried and changed the code where applicable to use the new property.

Migration code has been refactored and a new migration was added to address this change in identifiers. A change had to be made to the `store` method in order to keep user data from being lost with the migration.

I have done extensive manual testing to make sure nothing broke, but we should probably be thinking about writing automated tests for migrations as outlined in #329.

### DIFF
--- a/ConfCore/Event.swift
+++ b/ConfCore/Event.swift
@@ -19,7 +19,13 @@ public class Event: Object {
     @objc public dynamic var name = ""
 
     /// When the event starts
-    @objc public dynamic var startDate = Date.distantPast
+    @objc public dynamic var startDate = Date.distantPast {
+        didSet {
+            year = Calendar.current.component(.year, from: startDate)
+        }
+    }
+
+    @objc public dynamic var year = -1
 
     /// When the event ends
     @objc public dynamic var endDate = Date.distantPast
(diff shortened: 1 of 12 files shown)


insidegui:
NOTE: We should probably merge this before #367 and rebase/edit that one to reflect this change.

bcmn:
Oooh, one small hesitation -- how do we feel about line 44 of SessionsJSONAdapter?

`
let eventYear = eventIdentifier.replacingOccurrences(of: "wwdc", with: "")
`

insidegui:
Good catch, we should get that value from something else, I'll work on it this weekend.

insidegui:
@bcmn I was looking into this line you mentioned, it's used to populate the `year` property on `SessionAsset`, which is used to generate its unique identifier in the format `"year@sessionId~assetType"` (this happens in `SessionAsset.swift` on line 87). I'm pretty sure the live streaming machinery uses this `year` property to match a live video entity to a session (again, I should've documented this better).  Given that sessions outside of WWDC are not live streamed, and I don't see the `year` property being used anywhere else, I'll leave this part as-is for now and look into it more carefully in the future. I think we're too close to this year's event to change something critical related to live streaming.

Issue #364, opened 2018-04-22
Adopt Codable instead of using third party library


otaviolima:
- [ ] Remove  SwiftyJSON dependency by adopting Codable

insidegui:
Closed by #530

insidegui:
Thanks for the work on this @otaviolima and @allenhumphreys

PR #372, opened 2018-05-14
[WIP] Implement codable

otaviolima:
That is just a preview of the work.

TODO:

- [ ] Remove SwiftyJSON dependency from Tests
- [ ] Remove SwiftyJSON dependency from Preferences
- [ ] Remove SwiftyJSON dependency from CartFile
- [x] Review the implementation of the Codable for live videos, seems to be broken.
- [x] Rebase with currrent master

Resolves: #364


ConfCore/Adapter.swift
```diff
@@ -7,46 +7,36 @@
...
     }
+}
 
+extension KeyedDecodingContainer {
+
+    /// Utility used to provide a default value for missing keys but maintaining the behavior of throwing on data mismatch
+    public func decodeIfPresent<T>(_ type: T.Type, forKey key: KeyedDecodingContainer.Key, default: T) throws -> T where T: Codable {
+        return try self.decodeIfPresent(T.self, forKey: key) ?? `default`
+    }
 }
```
otaviolima: I might delete this as it is not strictly necessary.

ConfCore/Adapter.swift
```diff
@@ -7,46 +7,28 @@
...
 
 import Foundation
-import SwiftyJSON
 
-enum AdapterError: Error {
-    case invalidData
-    case unsupported
-    case missingKey(JSONSubscriptType)
+// This is needed to keep the same behavior that adapters had before
+// Where it could adapt array of items even if some of the individual items failed to adapt
```
allenhumphreys: @otaviolima I understand that you're keeping original code behavior, but I'm wondering whether that behavior is truly needed?

ConfCore/SessionsResponse.swift
```diff
@@ -8,11 +8,49 @@
...
+    private enum CodingKeys: String, CodingKey {
+        case events, sessions
+    }
+
+    public init(from decoder: Decoder) throws {
+        let context = DecodingError.Context(codingPath: decoder.codingPath,
+                                            debugDescription: "SessionResponse decoding is not currently supported")
+        throw DecodingError.dataCorrupted(context)
+
+//        let container = try decoder.container(keyedBy: CodingKeys.self)
```
allenhumphreys: Do we need this commented code?

ConfCore/Track.swift
```diff
@@ -43,22 +43,32 @@ public class Track: Object {
...
+    private enum CodingKeys: String, CodingKey {
+        case name, color, darkColor, titleColor, lightBGColor, ordinal
+        case identifier = "id"
     }
 
+    public convenience required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        let identifier = try container.decode(Int.self, forKey: .identifier)
+        let name = try container.decode(String.self, forKey: .name)
```
allenhumphreys: On another thing I was doing, I got tired of manually providing type information, so I wrote this:
```
extension KeyedDecodingContainer {

    func decode<T: Decodable>(forKey key: KeyedDecodingContainer.Key) throws -> T {
        return try decode(T.self, forKey: key)
    }

    func decodeIfPresent<T: Decodable>(forKey key: KeyedDecodingContainer.Key) throws -> T? {
        return try decodeIfPresent(T.self, forKey: key)
    }
}
```

Which then allows you to do `name = try container.decode(forKey: .name)`.

Also, not creating an intermediate variable (for simple decoding) would make the code more compact.

ConfCore/AppleAPIClient.swift
```diff
@@ -36,53 +35,45 @@ public final class AppleAPIClient {
         }
     }
 
-    private let jsonParser = ResponseContentTransformer { JSON($0.content as AnyObject) }
+
```
allenhumphreys: Vertical whitespace violation.

ConfCore/Session.swift
```diff
@@ -162,4 +162,140 @@ public class Session: Object {
...
+
+                self.assets.append(webPageAsset)
+            }
+        }
+
+        if let relatedContainer = try? sessionContainer.nestedContainer(keyedBy: RelatedCodingKeys.self, forKey: .related) {
+            if let resources = try relatedContainer.decodeIfPresent(FailableItemArrayWrapper<UnknownRelatedResource>.self, forKey: .resources)?.items.map({ $0.resource }) {
+                self.related.append(contentsOf: resources)
+            }
+
```
allenhumphreys: Trailing whitespace violation 😁😝

ConfCore/Adapter.swift
```diff
@@ -7,46 +7,28 @@
...
+struct FailableItemArrayWrapper<T: Decodable>: Decodable {
```
allenhumphreys: From a naming perspective, this is actually a wrapper that can never fail to decode.

ConfCore/Adapter.swift
```diff
@@ -7,46 +7,28 @@
...
-            switch itemResult {
-            case .success(let resultingItem):
-                return resultingItem
-            default: return nil
+        while !container.isAtEnd {
+            if let item = try? container.decode(T.self) {
+                items.append(item)
+            } else {
+                // container.currentIndex is not incremented unless something is decoded
+                _ = try? container.decode(Empty.self)
```
allenhumphreys: Any need for logging in this area, i.e. down an unexpected execution path?

ConfCore/FeaturedContent.swift
```diff
@@ -24,4 +24,30 @@ public class FeaturedContent: Object {
...
+    public required convenience init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        let sessionId = try container.decode(String.self, forKey: .sessionId)
+
+        self.init()
+
+        self.sessionId = sessionId
+
+        if let encodedEssay = try? container.decode(String.self, forKey: .essay) {
```
allenhumphreys: Why `try?` here instead of `decodeIfPresent`?

ConfCore/Session.swift
```diff
@@ -162,4 +162,140 @@ public class Session: Object {
...
+                webPageAsset.rawAssetType = SessionAssetType.webpage.rawValue
+                webPageAsset.remoteURL = permalink
+                webPageAsset.year = Int(eventYear) ?? -1
+                webPageAsset.sessionId = id
+
+                self.assets.append(webPageAsset)
+            }
+        }
+
+        if let relatedContainer = try? sessionContainer.nestedContainer(keyedBy: RelatedCodingKeys.self, forKey: .related) {
```
allenhumphreys: Why `try?`

ConfCore/Session.swift
```diff
@@ -162,4 +162,140 @@ public class Session: Object {
...
+        let title = try sessionContainer.decode(String.self, forKey: .title)
+        let summary = try sessionContainer.decode(String.self, forKey: .description)
+        let trackIdentifier = try sessionContainer.decode(Int.self, forKey: .track)
+        let eventContentId = try sessionContainer.decode(Int.self, forKey: .eventContentId)
+
+        self.init()
+
+        var mediaDuration: Double?
+
+        if let assetContainer = try? sessionContainer.nestedContainer(keyedBy: AssetCodingKeys.self, forKey: .media) {
```
allenhumphreys: Why `try?`

I'm asking about all of them because I don't want code that is so safe that it covers up bugs. Of course I understand if we're expecting failures and thus the `try?` is appropriate. Alternatively, you could do a do { } and then catch/ignore the "expected" errors and rethrow unexpected errors.

allenhumphreys:
Overall, this is great, a big win I think. 🎉 It clears several of the cyclomatic complexity warnings (which are really bugging me) as the adapters were some of the most complex functions in the app in terms of control flow.

 If you're feeling up for it @otaviolima, you could profile the implementations to see how long each takes to parse the data!

ConfCore/Session.swift
allenhumphreys: As an example, here is my implementation of Codable for RealmOptional
```
extension RealmOptional: Encodable where Value: Encodable {

    public func encode(to encoder: Encoder) throws {

        var container = encoder.singleValueContainer()

        try value.flatMap { try container.encode($0) } ?? container.encodeNil()
    }
}

extension RealmOptional: Decodable where Value: Decodable {

    public convenience init(from decoder: Decoder) throws {
        self.init()

        let container = try decoder.singleValueContainer()

        do {
            value = try container.decode(Optional<Value>.self)
        } catch let error as DecodingError {
            switch error {
            case .valueNotFound, .keyNotFound:
                return
            default:
                throw error
            }
        }
    }
}
```

insidegui:
Thanks for the thorough review! I agree that we should avoid `try?` because it can hide decoding bugs and would make it really difficult to debug when there are changes to Apple’s API.

ConfCore/Bookmark.swift
```diff
@@ -57,4 +57,33 @@ public final class Bookmark: Object, HasCloudKitFields, SoftDeletable {
...
+    }
+
+    public required convenience init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        let body = try container.decode(String.self, forKey: .body)
+        let timecode = try container.decode(Double.self, forKey: .timecode)
+        let duration = try container.decode(Double.self, forKey: .duration)
+
+        self.init()
```
allenhumphreys: @insidegui I'm curious what your thoughts are on setting the decoding up like this. I see that the purpose is to avoid object initialization until it has been determined that an object is needed, i.e. decoding succeeded, but in my opinion it seems to be optimizing for what * should * be the less likely scenario (most of the objects will decode successfully) and makes the code a bit longer and unwieldy.

insidegui: I'm sorry, I don't know what exactly it is you're referring to, to me this just looks like a normal custom `init` for a `Decodable` type 😅

allenhumphreys: Compared to this https://github.com/insidegui/WWDC/pull/530/files#diff-480a5a40eae0b47ad936d50c14cd5346R67

I rewrote it so `self.init` is called first, which means if the decoding fails, memory will be allocated and then deallocated but there is no need for the local variables

insidegui: Oh, I see. Yeah, I don't see any problem with that.

allenhumphreys:
Replaced by #372

PR #377, opened 2018-05-18
Implemented UI for related sessions

insidegui:
This is the first part of the UI side for #367 (issue #331). I've implemented the collection of related sessions at the bottom of the session details following @VicenteBorrell's design.

Some details on how the app decides which sessions to show:
- It only shows labs if their start time is greater than or equal to the current time (there's no need to show past labs)
- Items have to be double-clicked to go to the related session, single-click was too easy to select by accident
- When an item is selected in the videos tab that's not available in that tab (for instance, a lab), the app will switch to the schedule tab with the item selected

Even though this is only the related sessions portion of the implementation, it can already be merged. I'm going to implement the other related resources (documentation, sample code, etc) in another branch.

screen shot 2018-05-17 at 22 52 36


insidegui:
I tried leaving the scroll bar in. It looked ugly to me. 

bcmn:
Hm. Maybe we'd be getting into more bespoke customisations to achieve what I was picturing. But I agree, whilst it overlaps the cells, the scrollbar looks bad. I tried to mock this up very quickly, & it's all visually weighted & I didn't scale up the playout window to fill the top half of the space properly, but here's a rough mockup of what I was thinking:

screen shot 2018-05-20 at 17 33 34

insidegui:
Sounds like too much work for too little return 😅 I feel like most Mac users would know they can scroll as soon as they notice some of the content is clipped.

bcmn:
Fair enough. ¯\_(ツ)_/¯ Maybe it'll be a stretch goal PR I'll tinker with after the event 😉 

I'd prefer reverting the last two commits then, but I do definitely think single-click works way better. 🙂

insidegui:
> I'd prefer reverting the last two commits then, but I do definitely think single-click works way better.

Done 😉

PR #380, opened 2018-05-26
Implement sync for favorites, bookmarks and session progress

insidegui:
This is the initial implementation for CloudKit syncing of favorites, bookmarks and session progress. I decided to launch this feature as a beta first because we can't know for sure how it's working without having a decent amount of people actually using it. 

The feature comes disabled by default and can be enabled in preferences with a new switch I added.

screen shot 2018-05-26 at 00 36 30


insidegui:
Tests failed in CI because the initializer for `UserDataSyncEngine` was trying to initialize a `CKContainer` and that throws an exception when the app doesn't have the iCloud entitlement (the CI builds uses the non-iCloud scheme). I made sure to enclose every access to the `UserDataSyncEngine` and the class itself in `#if ICLOUD`.

insidegui:
Rebased, waiting for Travis to merge 😄

PR #384, opened 2018-05-26
Implemented session action links

insidegui:
Some labs and special events have action links associated with them to get directions to a place or schedule an appointment. This adds support for that, displaying the link in the summary view footer.

screen shot 2018-05-26 at 13 51 27

screen shot 2018-05-26 at 13 52 26

### DIFF
--- a/ConfCore/SessionInstance.swift
+++ b/ConfCore/SessionInstance.swift
@@ -98,6 +98,10 @@ public class SessionInstance: Object {
     /// See https://developer.apple.com/reference/eventkit/ekevent/1507437-eventidentifier
     @objc public dynamic var calendarEventIdentifier = ""
 
+    // Action link
+    @objc public dynamic var actionLinkPrompt: String?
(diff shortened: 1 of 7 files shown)


insidegui:
You're totally right, I actually intended to only show them for instances that start after the current date and time and forgot. Will do and update the PR.

insidegui:
@bcmn Updated to only show up when the session starts at a later date 😉

bcmn:
Looking good. I did notice the link data doesn't get get parsed if using an existing database. This shouldn't be an issue as far as I can tell, but I thought I'd mention it to see if you can think of any scenario in which it gets weird? 🤔

insidegui:
Oh, I forgot to add the code to merge that in to existing data. It shouldn't be a big issue for now since we only show them for future sessions.

bcmn:
Yeah, that's what I figured ¯\_(ツ)_/¯

PR #387, opened 2018-05-28
Check for explicit field editors

davedelong:
The problem with the code is that it's only checking a single window for an `NSTextView`. In the case (for example) when the player has popped out to the PIP window, then weird things happen.

So instead, we'll simply check all windows to see if any of them has an active field editor. If one does, then we'll defer and let that field editor handle the space. If no window is actively editing text (ie, there is no field editor), then we'll process the space key as the play/pause toggle.


davedelong:
This fixes #227

PR #392, opened 2018-05-29
Implemented "Lab by Appointment" session type

insidegui:
This also fixes an issue with thumbails showing up all back in the schedule view.

### DIFF
--- a/ConfCore/SessionInstance.swift
+++ b/ConfCore/SessionInstance.swift
@@ -10,7 +10,7 @@ import Cocoa
 import RealmSwift
 
 public enum SessionInstanceType: Int {
-    case session, lab, video, getTogether, specialEvent
+    case session, lab, labByAppointment, video, getTogether, specialEvent
 
     init?(rawSessionType: String) {
         switch rawSessionType {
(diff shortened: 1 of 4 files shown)


ConfCore/SessionInstance.swift
```diff
@@ -10,7 +10,7 @@ import Cocoa
 import RealmSwift
 
 public enum SessionInstanceType: Int {
-    case session, lab, video, getTogether, specialEvent
```
allenhumphreys: I think you'll want to add it to the end, because the type is stored in a Realm as the `rawValue`, which means for existing db's all the sessions would theoretically get new types.

insidegui: I don't think that's an issue since when merging the content from the feed, the types will be overridden by whatever comes from the feed. I'll do that anyway just to be extra careful.

WWDC/SessionCellView.swift
```diff
@@ -75,7 +75,7 @@ final class SessionCellView: NSView {
             self?.imageDownloadOperation?.cancel()
 
             self?.imageDownloadOperation = ImageDownloadCenter.shared.downloadImage(from: imageUrl, thumbnailHeight: Constants.thumbnailHeight) { [weak self] url, _, thumb in
-                guard url == imageUrl else { return }
+                guard url == imageUrl, thumb != nil else { return }
```
allenhumphreys: Nice find, one day the image downloading will have 0 bugs... one day.

PR #394, opened 2018-05-30
Fixes some layout and scrolling issues on playlists

allenhumphreys:
Addresses:
1. Vertical scroller would disappear when set to always visible
2. Collection views would get messed up if scrollers set to always visible
3. Bottom padding of page caused scroller to not reach bottom
4. Collection views weren't scrollable until SOMETHING happens, such as window resize or contents get changed
5. Outer scroll view bounced horizontally if your mouse wasn't over a collection view


insidegui:
Good job!

PR #406, opened 2018-05-31
Modify related content layout to allow scroller

allenhumphreys:
This is a fix and an enhancement.

It fixes the layout when a users system preference is to always show scroll bars. It does this by forcing overlay scrollers, which breaks expectations but simplifies layout issues considerably.

The enhancement is making room for the overlay scroller so that it doesn't overlay the content. The scroller flashing is default behavior.

relatedsessionsscroll


insidegui:
>This is a fix and an enhancement. ...

@allenhumphreys Nice!

Issue #409, opened 2018-06-02
Format menu shouldn't exists


JoelNichols:
The format menu shouldn't exist. Just about everything it offers is greyed out anyway and it just looks sloppy that it's been left in when none of the options are relevant to the app.

bcmn:
☝🏻 That message only gives half the story -- see #410 for more details.

PR #410, opened 2018-06-02
Removes Format from menu bar

bcmn:
As noted in #409, the Format menu serves little (no?) purpose. This PR removes it entirely.


bcmn:
Updated after discussion. We intend to support the Format menu for annotations at a later date -- therefore, we'll keep the code, but hide the menu until needed. While at it, we've updated the help menu item to direct to the issues page of this repo.

PR #415, opened 2018-06-05
More live streaming improvements

insidegui:
Using `Calendar` for time calculations, added some extra logging to improve debugging and hooked up Cmd+R refresh action to also refresh live sessions.

### DIFF
--- a/WWDC/Constants.swift
+++ b/WWDC/Constants.swift
@@ -20,7 +20,10 @@ struct Constants {
     static let watchedVideoRelativePosition: Double = 0.97
 
     /// How many seconds between each live session check
-    static let liveSessionCheckInterval: TimeInterval = 10.0
+    static let liveSessionCheckInterval: TimeInterval = 60.0
+
+    /// How many seconds of tolerance to give the live session check timer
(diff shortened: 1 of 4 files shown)


WWDC/Constants.swift
```diff
@@ -22,8 +22,8 @@ struct Constants {
     /// How many seconds between each live session check
     static let liveSessionCheckInterval: TimeInterval = 10.0
 
-    /// How many seconds to subtract from the start time of a live session to consider it live
-    static let liveSessionStartTimeTolerance: TimeInterval = 300
+    /// How many MINUTES to subtract from the start time of a live session to consider it live
```
allenhumphreys: “to *add* to the start time” 😇

WWDC/LiveObserver.swift
```diff
@@ -70,27 +70,35 @@ final class LiveObserver {
     }
 
     func refresh() {
-        specialEventsObserver.fetch()
+        NSObject.cancelPreviousPerformRequests(withTarget: self)
```
allenhumphreys: This is a good change, using delorean I could start live sessions then they would be stuck because the live observer never started to clear the live flag. Now this should give a way to override it.

PR #429, opened 2018-06-06
Remove duplicated tooltips

pedrovieira:
The tooltips from the "Skip Back and Forward by 30 Seconds" toggle and label were also, incorrectly, used in the "Enable iCloud syncing (beta)" toggle and label. I decided to just delete them instead of writing new tooltips because the iCloud pref already has a descriptive label associated with it.


insidegui:
Nice catch!

PR #432, opened 2018-06-07
Implements 431, copy URL to clipboard

allenhumphreys:
Does anyone know if there is a system asset that can be used for this share action?

screen shot 2018-06-07 at 10 33 30 am


insidegui:
@allenhumphreys You can use this one
[copy.pdf](https://github.com/insidegui/WWDC/files/2081151/copy.pdf)

PR #439, opened 2018-06-08
Fixes session lists to allow arbitrary navigation

### LINKED ISSUE #395: Playlist and related linking doesn't always show the linked session in the destination tab
- Version of the app
 * master

Choosing a video from a playlist/related content will always show the correct tab, but

1. If the target video/session is not in the list due to an active filter
2. If the tab has never been populated since app launch

the video will not be selected in the list.

allenhumphreys:
fixes #395

It clears active filters if they're preventing navigation. 
Provides an abstraction of delivering filter queries and allows the session table view to be a little more involved. I did this for flexibility when updating the table, though I'm not currently taken advantage of the flexibility it affords because I initially started down this path thinking I would be forcing items into the list that are navigated to. The solution leaves the responsibility of creating the search predicate in the SearchCooridinator, which is good I think. I also reorganized the code to make the file a little more approachable by moving supporting stuff into a separate file!

I can no longer test this because my featured content has stopped populating! Please help @insidegui

### DIFF
--- a/ConfCore/SyncEngine.swift
+++ b/ConfCore/SyncEngine.swift
@@ -61,9 +61,9 @@ public final class SyncEngine {
                 self.storage.store(contentResult: scheduleResult) { error in
                     NotificationCenter.default.post(name: .SyncEngineDidSyncSessionsAndSchedule, object: error)
 
-//                    guard error == nil else { return }
+                    guard error == nil else { return }
 
-//                    self.syncFeaturedSections()
+                    self.syncFeaturedSections()
                 }
             }
         }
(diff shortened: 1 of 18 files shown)


bcmn:
Hm, I’m also unable to populate the Featured tab. It seems the endpoint is responding fine to my Paw requests, I’m not sure what’s going on without more exploration.

allenhumphreys:
This is ready for proper review now, I think.

allenhumphreys:
Would just need to tweak the layout to make room for them, they currently overlap the bottom text if you unhide them.

PR #477, opened 2018-06-19
Addresses player controls hiding while still preventing disappearing cursor

allenhumphreys:
Window management is hard 😔


insidegui:
Working perfectly, I even tried it with an external display 😄

PR #480, opened 2018-06-20
Ensure player (un)hides correctly on selection changes, fixes #479

### LINKED ISSUE #479: Player fails to hide, obscuring the image/play button of selected sessions
Introduced by #428 , we're trying to achieve the correct balance of player hiding.

This bug results in the player being visible on sessions other than the currently playing session.

allenhumphreys:
I removed the `canRestorePlaybackContext` variable as I couldn't tell if it had a true purpose. The way it was being used seemed to not align with its name.

I'm not sure if this was a feature or a bug, but.. If you exited PiP while the playing session was selected, then the next time you selected a different session, auto-PiP would not work. This seemed to be implemented by way of the `canRestorePlaybackContext` flag, but based on the flags name, it seemed unintentional.

## Application Flows/Test Cases

note: Expectations in square brackets

### PiP

note: Unless otherwise obvious, the player should be playing

1. Enter pip
[The player view shows the "Playing in Picture in Picture" view]
2. Return from pip
[Player returns to main window].  
###
1. Enter PiP
2. Choose X to exit
[Playback should stop and player should return to main window]
###
1. Play session
2. Pause session
3. Change selection/tab
[Player should remain in the window, new selection should be visible]
###
1. Enter pip
2. Change session selection
[the player should hide, revealing the newly selected session]
3. Return from pip
[The playing session's row should be selected and the player should return in a visible state]
###
1. Enter pip
2. Change session selection on current tab
[the player should hide, revealing the newly selected session]
3. Change tab
3. Return from pip
[The playing session's tab and row should be selected and the player should return in a visible state]
###
1. Play session
2. Change selection or tab
[PiP should be activated automatically]
###
1. Enter PiP
2. Change selection or tab
3. Choose X to exit
[Playback should stop and the main window should remain in its current state]
4. Return to session you played
[Player should be visible]
###
1. Play session
2. Pause session
3. Select different session
4. Use remote control, touch bar, now playing, etc to resume playback
[Playback should resume, but the main window should remain as is]
5. Reselect playing session
[The playing session should be revealed with the player visible]
###
1. Enter pip
2. Select new session
3. Play new session
[The currently playing video should change in the PiP window and the shelf should show the "Playing in Picture in Picture" view]
###
1. Have both videos and schedule on the same session
2. Play the session on one tab
3. Switch to the other
[Auto pip activates]
###
1. Have both videos and schedule on the same session
2. Enter PiP from tab A
3. Switch to tab B (with the playing session selected)
4. Hit "return from pip" button
[Should switch to tab A with video player visible]

### Fullscreen

1. Enter fullscreen
2. Switch back to main window space
3. Change selection
4. Switch back to FS space
[Player should be visible]
5a. Exit full screen via X(red) button
[Playback should stop, main window should reappear without changing selection
5b. Exit full screen via zoom(green) button
[Playback should container, main window should reappear and select the currently playing session]
###
1. Enter fullscreen
2. Switch back to main window space
3. Change selection
4. Play new session
[Playback should change and the main window should remain the same]
5. Switch back to FS space
[Should see the newly played session playing here]


allenhumphreys:
There is an view updating artifact that is noticeable when switching between sessions when one of them is the playing session. I animated the (un)hiding of the player to soften it a little, but in the future I'd like bring the transition into one smooth change.

insidegui:
On this branch, when playing a video in fullscreen, if I select another session, the fullscreen window player disappears and exiting fullscreen doesn't bring the playing session back.

allenhumphreys:
Thanks for checking it. There’s so many usage permutations! I’ll reopen when I have another minute to keep investigating.

allenhumphreys:
@insidegui This should be fixed now. Only scenario I haven't tested on is a multi-monitor full screen situation.

insidegui:
Will test. Thanks

insidegui:
With the updated branch, if I enter fullscreen, select another video on the list and exit fullscreen, the player gets attached to the session I have selected, not the session it belongs to.

allenhumphreys:
@insidegui That particular bug was already introduced as part of my initial fix for the player hiding while in full screen. I think the final solution to all the hiding mess is to put the player in a container in the shelf and then hide the container instead of the player itself. This completely disconnects the hiding when the player is in full screen.

Per the Slack discussion, exiting fullscreen (via esc key or the green button) will jump the UI back to the playing session. Closing the window will not. Much like the PiP does today. Additionally, when the player is fullscreen, playing a new session will not exit the fullscreen. This is probably great for multi-monitor users.

Its funny that all of this came about because the player was hiding while in full screen 💯

insidegui:
@allenhumphreys Working great now! Can you fix the conflict and merge? Thanks!

allenhumphreys:
There is one last teensy tiny bug I want to understand. It hast to do with playing a video that is on both tabs.

PR #485, opened 2018-06-25
Improves players handling of information loading to prevent hangs

### LINKED ISSUE #465: App hangs when starting video playback
This will be worse/more noticeable when you have a slow connection. For the time being, its best to wait for for the video to start playing back before interacting with the app anymore.

There is an inadvertent synchronous, yet network dependent, call being made when starting the play. (Possibly more than 1).

The main one of concern seems to be:
```
player.addObserver(self, forKeyPath: #keyPath(AVPlayer.currentItem.currentMediaSelection), options: [.initial, .new], context: nil)
```

The option to get an initial value from KVO is the same thing as calling the method path directly, so this becomes a blocking call.

We need to do some investigation to make sure all information gathered is being handled in appropriately asynchronous ways (the player volume, for instance, does not have to be handled in a special way).

And make sure the loading spinner works! :-)

allenhumphreys:
...especially for users on slower internet connections

Fixes #465

I also put in some very basic error handling for displaying AVFoundation errors and subsequently reset the player after the error state is entered. When changing videos, the player resets some state to avoid having the controls show information from the previous video while loading the new one. The loading spinner will show after a brief delay, otherwise it may appear to flash on then off. The length of the delay is up for debate.

I'd like feedback on my approach and a little help testing.

### DIFF
--- a/PlayerUI/Util/AVAsset+AsyncHelpers.swift
+++ b/PlayerUI/Util/AVAsset+AsyncHelpers.swift
@@ -1,0 +1,22 @@
+//
+//  AVAsset+AsyncHelpers.swift
+//  PlayerUI
+//
+//  Created by Allen Humphreys on 6/25/18.
+//  Copyright © 2018 Guilherme Rambo. All rights reserved.
+//
+
+import AVFoundation
+
+extension AVAsset {
+
+    public var durationIfLoaded: CMTime? {
+
+        let error: NSErrorPointer = nil
+        let durationStatus = statusOfValue(forKey: "duration", error: error)
+
+        guard durationStatus == .loaded, error?.pointee == nil else { return nil }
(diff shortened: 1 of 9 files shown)


insidegui:
This is a huge improvement 👏

I noticed only one issue: the subtitles menu is not showing for me when I click the subtitles button.

allenhumphreys:
>>>  You determine which alternative media presentations are available by asynchronously loading and calling the asset’s availableMediaCharacteristicsWithMediaSelectionOptions property, which returns an array of strings indicating the available media characteristics. The possible values returned are audible (alternative audio tracks), visual (alternative video tracks), and legible (subtitles and closed captions).

Turns out the warning from the async KVO system was worth paying attention to :-P

PR #488, opened 2018-06-28
Optimizes drawing of WWDCTableRowView

allenhumphreys:
Now, when you're playing a video, the table view animations are smooth. You may also notice an improvement in scrolling. And even updates when there is no video playing. This has taken me so long to figure out 🎉😅😭


insidegui:
Awesome 👏

Issue #458, opened 2018-06-12
Speed isn't maintained after switching videos.


yarnichat:
1. Play a video and change the speed to, say, 2x.
2. Play a different video.
3. The speed button still says 2x, but the video plays at 1x (!).

The speed setting should be preserved when switching between videos.

the7thgoldrunner:
If we still want to fix this, I think I'll take it.

allenhumphreys:
@bcmn Any issue with that?

bcmn:
Yeah that’s fair enough, I’ve had a lot of things come up recently & not
had time to fully update this stuff to unassign myself and everything. My
bad.

PR #496, opened 2018-07-21
Bug fix (#458): Playback Speed reverts to 1x after switching videos

the7thgoldrunner:
The root of the issue is that between videos, the internal AVPlayer instance is recreated to stream/play the new video, and when this happens, the playback speed state isn't properly restored. There are a couple of ways to go about fixing this issue, we picked here what seemed the most architecturally-sound.

### DIFF
--- a/WWDC/VideoPlayerViewController.swift
+++ b/WWDC/VideoPlayerViewController.swift
@@ -143,9 +143,8 @@ final class VideoPlayerViewController: NSViewController {
 
         setupPlayerObservers()
 
-        player.play()
-
         playerView.player = player
+        playerView.play(self)
 
         if let playbackViewModel = playbackViewModel {
             playerView.remoteMediaUrl = playbackViewModel.remoteMediaURL


the7thgoldrunner:
Sounds good, I'll be checking in case any changes are needed.

allenhumphreys:
Nice fix, super clean. Definitely looks like this was how it was intended to be used. 

I thought of a follow up, however, that could prevent something like this from happening in the future. `PUIPlayerView` observes the rate property of its player, and updates it's UI accordingly. What if we logged a message in the event that a non-zero rate appears that doesn't match with the UI rate?

the7thgoldrunner:
We can do that. Should I add it here or should it be on a different pull request?

allenhumphreys:
Here is fine.

the7thgoldrunner:
@allenhumphreys looking at the code, I'm not sure logging the disparity would be of any use. Whenever `PUPlayerView`'s rate observer is called, we already update the UI to match the speed of the current `playbackSpeed` property - the bug was caused because the observer wasn't called on the default 1.0x speed of new `AVPlayer` instances. 

I'm happy to do the change if you still want me to, but in the cases where (like this bug) the observer isn't called for a change in playback speed, the observer won't notify us about the disparity, and even if it did, the fact that the observer gets called means we're updating the UI to reflect the current playback speed.

allenhumphreys:
@the7thgoldrunner The observer does get called when the player is changed. I'd argue that the bug is caused by the fact that we don't update the rate button when the rate changes. See [this line](https://github.com/insidegui/WWDC/blob/master/PlayerUI/Views/PUIPlayerView.swift#L292) to see what I mean. It's not a huge deal though, since Gui merged the PR already we can let it go. I have a tendency to program too defensively, sometimes.

the7thgoldrunner:
Alright. Letting it go.

Issue #422, opened 2018-06-06
Downloads window


mackuba:
I really miss the Downloads window from the old version (to which I even submitted some PRs, I think)... I would love to see how many downloads I have started and how they're going, and maybe pause some of them to prioritize the one I want to see right now, instead of having to switch back and forth between session pages to check the circular progress bars there.

If, hypothetically speaking, someone were to send you a PR with this ;) would you be interested? What was the reason for removing it during the redesign? Would it make more sense as a small window as before, accessed from the Window menu, or rather as something inside the main window somewhere?

insidegui:
We didn’t have time to reimplement it for the new version yet, but that’s planned. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

AnalyzePlatypus:
This would be great.

VicenteBorrell:
Let me know if you need the design for this 👍

insidegui:
@VicenteBorrell Yes, please :) I think we could show an icon in the toolbar when there are active downloads and show the list of downloads in a popover or window when that icon is selected.

allenhumphreys:
![41206696-4c1776bc-6d08-11e8-92d6-89bae50ac173](https://user-images.githubusercontent.com/8225090/41206822-33b630be-6cd0-11e8-82ba-c012ef440910.png)

Thanks to @mackuba

bcmn:
Evoking the Safari download manager sounds perfect to me.

insidegui:
> Evoking the Safari download manager sounds perfect to me.

Same.

insidegui:
Having ETA and notifications for completion would be nice.

VicenteBorrell:
Are we going to adopt Dark Mode? Should I follow it? 😊

allenhumphreys:
Work began [here](https://github.com/insidegui/WWDC/pull/497)

PR #497, opened 2018-07-23
Refactor titlebar, stubs for many components of downloads window

allenhumphreys:
Begins work on #422, starts by refactoring the tab bar controller to use a custom UI rather than hijacking NSToolbar. General architectural stubs are in place, with a dummy button appearing for debug builds. It looks like I might be winding down my work for a little while so I thought this was a sensible stopping point that gives a lot of direction in case anyone wants to pick up the work in the meantime.


insidegui:
Looks good, thanks for getting this started

PR #500, opened 2018-08-10
[gardening] Update dependencies, cleanup asset filtering

allenhumphreys:
Just taking a minute to review our dependencies, and I noticed how many places "rawAssetType == %@" was written in the app.

Of concern, swift-protobuf made a change that will cause [unknown JSON fields to cause an error](https://github.com/apple/swift-protobuf/pull/771). I don't know if this affects ChromeCastCore in practice since I don't have a compatible chrome cast device.

SwiftLint fixed their silly fallthrough rule, they replaced it with something that only disallows cases that only have a fallthrough statement.


allenhumphreys:
@insidegui Would you like me to update ChromeCastCore as well?

insidegui:
@allenhumphreys Not yet, I still have to test the fix better and there are other changes that need to be done.

PR #501, opened 2018-08-23
Support filtering on bookmarked sessions, some UI fixes for bookmarks in timeline

allenhumphreys:
Some general love to the annotation UI.

• Clicking anywhere else in the app will close the annotation window
• The white corners are no longer visible in the annotation window
• The annotation bubble highlighting is correctly removed after editing an annotation
• No longer receive bell when using escape or cmd + enter to exit an annotation
• Ability to filter by sessions with bookmarks on them

annotations

### DIFF
--- a/ConfCore/Storage.swift
+++ b/ConfCore/Storage.swift
@@ -201,7 +201,7 @@ public final class Storage {
 
             let instances = backgroundRealm.objects(SessionInstance.self).sorted(by: SessionInstance.standardSort)
 
-            var previousStartTime: Date? = nil
+            var previousStartTime: Date?
             for instance in instances {
                 guard instance.startTime != previousStartTime else { continue }
 
--- a/PlayerUI/Controllers/PUIAnnotationWindowController.swift
+++ b/PlayerUI/Controllers/PUIAnnotationWindowController.swift
@@ -20,6 +20,8 @@ class PUIAnnotationWindowController: NSWindowController {
 
         super.init(window: window)
 
+        window.isOpaque = false
+        window.backgroundColor = .clear
         window.contentView = vfxView
     }
 
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -221,9 +221,7 @@ public final class PUIPlayerView: NSView {
     private func setupPlayer() {
         elapsedTimeLabel.stringValue = elapsedTimeInitialValue
         remainingTimeLabel.stringValue = remainingTimeInitialValue
-        timelineView.playbackProgress = 0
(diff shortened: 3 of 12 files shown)


insidegui:
This is great, I've been willing to do those fixes for a long time. Any idea as to why the CI failed? It's late and I'm going to bad, I'll have another look tomorrow.

insidegui:
When you drag an annotation around and then release it, it opens up the annotation window, I think this should not happen (the window should only show up with a single, regular click).

Another issue is that when in fullscreen, pressing the esc key (which actually becomes an "exit from fullscreen" key in Touch Bar Macs) is exiting fullscreen instead of dismissing the annotation window. I think this could be solved by having the annotation window provide its own Touch Bar.

I'm pretty sure this is an issue that already existed before, but since you're working on this, maybe you can try to fix it :)

allenhumphreys:
I can't get the window to appear after dragging the annotation around! Maybe I'm misunderstanding your description.

As for the touch bar issue, I'd like to fix it by always making the annotation window key instead of what currently happens
```
if annotation.isEmpty {
    annotationWindow.makeKeyAndOrderFront(nil)
} else {
    annotationWindow.orderFront(nil)
}
```

This will mean the text view will always become the first responder and it provides it's own touch bar. Doing this would also prevent possible other edge cases. Thoughts?

insidegui:
Here's a gif showing the issue: 
issue


> This will mean the text view will always become the first responder and it provides it's own touch bar. Doing this would also prevent possible other edge cases. Thoughts?

Sounds good to me

allenhumphreys:
Ooohhh, you have to drop it back where it started! I kept dropping it in new places.

insidegui:
Also note the "overlapping timestamps from hell"

allenhumphreys:
Now that you mention the timestamps, I think we should really move the time remaining and playback progress times to somewhere else. Since all that avoidance stuff that was done for the seek preview completely does not happen for annotations. We could move them to the ends of the timeline or below it? (Not part of this PR though, I woke up with a cold and now my brain is fuzzy)

insidegui:
I really like the way the player looks with the labels where they are 😅

allenhumphreys:
I do, too, actually. Maybe we can make their views owned by the timeline UI rather than the player, and then hide them when mousing over the timeline so the seek time preview and annotations can take precedence?

insidegui:
That would make sense, yeah

insidegui:
To avoid possible layout issues, maybe the timeline could delegate the hiding of the labels 🤔

insidegui:
Something like `timelineViewTimePreviewLabelWillHitEdge(_ view: PUITimelineView)` and then the player can hide them

PR #503, opened 2018-09-12
Migrate to swift 4.2

### LINKED ISSUE #502: Migrate to Swift 4.2/Xcode 10
The GM is out, and [travis](https://docs.travis-ci.com/user/reference/osx/#os-x-version) finally learned their lesson and dropped the beta from their image identifiers so that the switch for them will be automatic.

allenhumphreys:
Fixes #502

Work in progress, wanted others to see the changes to the string identifiers in various places. `NSImage.Name` still exists but I didn't use it.

Per @insidegui 's suggestion, I put all the identifiers as statics on the identifier type, which is essentially `String` because they're all type aliases now. I ran into a name collision, which could be avoided by file private extensions. Just want to make sure this is the direction we want to go. Personally, my vote is to just use string literals at the call site like we would in UIKit.

This is also testing travis.


insidegui:
> Personally, my vote is to just use string literals at the call site like we would in UIKit

Let's do that

allenhumphreys:
I was able to get travis to build by updating realm. So now this PR is a realm update, too.

insidegui:
@allenhumphreys Is this safe to merge?

allenhumphreys:
I believe it is, if you’ve built it with CloudKit enabled. The new build system checks signing BEFORE building so I didn’t get to compile it with CK enabled. Any errors would be trivial renaming, though. 

Also, this would lead to the first release with the new build system, so there are some unknowns there.

I’m confident the changes here are correct and ready, though.

PlayerUI/Views/PUIPlayerView.swift
```diff
@@ -1020,14 +1020,14 @@ public final class PUIPlayerView: NSView {
 
         guard let durationTime = player.currentItem?.duration else { return }
 
-        let modifier = CMTimeMakeWithSeconds(seconds, durationTime.timescale)
+        let modifier = CMTimeMakeWithSeconds(seconds, preferredTimescale: durationTime.timescale)
```
sindresorhus: Slightly off-topic, but I would use the modern constructor instead:
[`CMTime(seconds: Double, preferredTimescale: CMTimeScale)`](https://developer.apple.com/documentation/coremedia/cmtime/1489786-init)

Issue #504, opened 2018-09-14
"Tech Talks" are incorrectly appended with a year in the TableView


bcmn:
Sessions belonging to the "Tech Talks", "Developer Insights" & "App Store Connect" events get incorrectly appended with a year in the TableView on the left, as shown below:

![image](https://user-images.githubusercontent.com/11646957/45539325-4a8c8000-b801-11e8-9a37-345546dd9c55.png)

Specifically, the issue is related to [this function](https://github.com/insidegui/WWDC/blob/master/WWDC/SessionViewModel.swift#L192) in `SessionViewModel.swift`:
```
    static func subtitle(from session: Session, at event: ConfCore.Event?) -> String {
        guard let event = event else { return "" }

        let year = Calendar.current.component(.year, from: event.startDate)

        // Currently, there's only one event which is not a WWDC (the "Fall 2017" event),
        // for some reason it includes the year on its name, while WWDC editions do not,
        // so we have to use this workaround to avoid displaying the year twice
        let name = event.name.replacingOccurrences(of: " \(year)", with: "")

        return "\(name) \(year) · Session \(session.number)"
    }
```

We only want to append the year to "WWDC" event names, without altering any other event names.

marcocapano:
I'm on it :) 😊

marcocapano:
PR submitted🎉

insidegui:
Thanks @marcocapano

marcocapano:
Since a PR is already on this I guess we could close it?

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bcmn:
Typically, we close issues once the PR is merged -- especially as there's a reference to the PR above. It's possible to use [Github's keywords](https://help.github.com/articles/closing-issues-using-keywords/) to make this even easier, & even do it automatically.

insidegui:
Implemented in #506

PR #506, opened 2018-09-14
Removed the year from the label for non-WWDC events

marcocapano:
From issue #504 : "We only want to append the year to "WWDC" event names, without altering any other event names."

This PR fixes this problem.
If you can think of any better solution I'm happy to edit this😊

### DIFF
--- a/WWDC/SessionViewModel.swift
+++ b/WWDC/SessionViewModel.swift
@@ -193,13 +193,12 @@ final class SessionViewModel {
         guard let event = event else { return "" }
 
         let year = Calendar.current.component(.year, from: event.startDate)
+        var name = event.name
 
-        // Currently, there's only one event which is not a WWDC (the "Fall 2017" event),
-        // for some reason it includes the year on its name, while WWDC editions do not,
-        // so we have to use this workaround to avoid displaying the year twice
-        var name = event.name.replacingOccurrences(of: " \(year)", with: "")
-
-        //We only want to add the year to WWDC event names.
+        /*
+        We want to make sure that WWDC events show the year
+        So we add it it not present.
+        */
         if name == "WWDC" {
             name.append(" \(year)")
         }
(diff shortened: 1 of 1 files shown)


bcmn:
Thanks @marcocapano ! I'm glad to see a new contributor. 😊 

One note, & maybe this was a lack of clarity with my original issue -- the logic is a little complicated: currently, it alters all strings that get passed in (line 200). We then go to an extra step of adding it back where necessary (the if statement on .

It might be simpler to no longer remove the year from _all_ strings. Instead, perhaps we should just check if the string only says "WWDC", & then append the year if that is true? Otherwise, we'd just return the original string, & assume that they don't need a year.

Also, the comment on line 197-199 isn't really true any more. Do you think we could remove it?

marcocapano:
Thanks @bcmn for the help!
I've made some changes based on your comment. Let me know what you think of it 😊

Also: In some other files there are some swift lint warnings. Can I include fixing some of them in the PR, since the contribution guidelines say that I can't make a separate PR only for that?

bcmn:
🎉 Looks great -- I think it'll be much clearer for someone new. Can you review this if you find the time, @insidegui ?

Regarding Swiftlint: if it's a small fix then maybe it can go in this PR, but if it's a big fix then I think a separate PR just for them is entirely reasonable, especially if they're refactoring more complex code. 
I think we're happy to merge this branch as soon as one of us has reviewed it in-app, so adding larger fixes might just slow that down!

marcocapano:
I’ll do those in a separate one then, thanks! 😊

allenhumphreys:
@marcocapno Welcome and thanks for jumping in to fix this one. Regarding the swiftlint warnings, most of them are going to be fixed by the pending PR that adopts codable. I fixed another in the Swift 4.2 PR. Just a heads up.

insidegui:
Tested and approved. Great job, @marcocapano !

PR #508, opened 2018-10-06
WIP: Mojave design update (Fixes #507)

### LINKED ISSUE #507: Mojave issues
I noticed these when linked against the 10.14 SDK. I'll be doing some work to get these fixed and adopt the standard dark mode on Mojave instead of doing our own.

screen shot 2018-10-05 at 20 34 27

insidegui:
This updates the project so the app looks better on Mojave's native dark mode.

screen shot 2018-10-05 at 21 16 49


allenhumphreys:
When the window is inactive, the filter button loses it's contents 
screen shot 2018-10-19 at 4 57 45 pm

insidegui:
> When the window is inactive, the filter button loses it's contents

Weird, it happens for me in `master` as well (on Mojave).

Other than that, any other issues we should fix before merging? I'd be happy to let that slide for now

allenhumphreys:
Not that I know of. Need to check the downloads window combined with this, though.

allenhumphreys:
screen shot 2018-10-23 at 6 30 56 pm

Found 1!

insidegui:
Good catch

insidegui:
Alright, implemented a custom progress indicator to adapt to the accent color a little better.

As discussed on Slack, we should probably add a pause/continue glyph to the indicator and change its behaviour from the current one (which just cancels the download). Those changes can be made in another PR.

Some samples of the new indicator in different colors:

screen shot 2018-10-24 at 18 11 49
screen shot 2018-10-24 at 18 11 56
screen shot 2018-10-24 at 18 12 06
screen shot 2018-10-24 at 18 12 13

allenhumphreys:
I love this, it looks great and better than the pie progress I think. It could use a track in the determinate state though. With a large file and slow internet you see nothing. Alternatively (or in addition to the track), it could have a minimum filled amount so it would seem like there is always a tiny bit of progress. These are nitpicks, though, and if you're busy we should move on!

insidegui:
[There is a track](https://github.com/insidegui/WWDC/pull/508/files#diff-b1df9431d09add9b9b40f4f0955933bfR156), but I think it's too faint, should be a little bit more opaque.

insidegui:
I think it's better now

image

allenhumphreys:
Yeah I couldn’t see the track on my computer when I tested it. Or in your screenshots. This looks better!

PR #512, opened 2018-10-21
Downloads management window, pause, resume, cancel, etc

### LINKED ISSUE #422: Downloads window

allenhumphreys:
Largely resolves #422 

I took some design license, but I'm open to feedback. The downloads window lets you manage your downloads. Downloads continue after quitting and can be accessed the next time the app is launched. The window design mimics Safari's implementation. 

This PR does *not*:
* provide an overall progress in the button
* provide time remaining estimates
* provide throughput information

This PR does:
* maintain all functionality that already exists
* attempts to work for the users
* attempts to be performant in many usage scenarios, including attempting to download the entire collection of WWDC videos
* Updates some tangentially related code for clarity
* Hopes to be loved and cherished by the users
* Has a major weakness if you cancel a large number of downloads while the management window is open


allenhumphreys:
downloadall

insidegui:
> including attempting to download the entire collection of WWDC videos

Savage 😛

> Has a major weakness if you cancel a large number of downloads while the management window is open

What happens? 👀

--

Code looks good, I'm going to try it out later and give you more feedback.

WWDC/DownloadManager.swift
```diff
@@ -12,69 +12,61 @@ import ConfCore
...
+    // MARK: - Session-based Public API
+
     func start(with storage: Storage) {
         self.storage = storage
 
         backgroundSession.getTasksWithCompletionHandler { _, _, pendingTasks in
             for task in pendingTasks {
-                if let key = task.originalRequest?.url!.absoluteString {
-                    self.downloadTasks[key] = task
+                if let key = task.originalRequest?.url!.absoluteString,
```
insidegui: I know `url` is probably never going to be nil, but just in case... ¯\_(ツ)_/¯ 

```suggestion
                if let key = task.originalRequest?.url?.absoluteString,
```

WWDC/DownloadManager.swift
```diff
@@ -87,96 +79,60 @@ final class DownloadManager: NSObject {
...
-    func pauseDownload(_ url: String) -> Bool {
-        if let task = downloadTasks[url] {
-            task.suspend()
-            NotificationCenter.default.post(name: .DownloadManagerDownloadPaused, object: url)
-            return true
+                for (url, download) in successfullyStartedTasks {
+                    download.task?.resume()
+                    NotificationCenter.default.post(name: .DownloadManagerDownloadStarted, object: url)
+                }
+            }
```
insidegui: It's not clear to me what the code above is doing, should probably add some comments.

allenhumphreys:
> What happens? 👀

Every time a download gets canceled the table reloads. If you have a lot of downloads in the table it makes for a very slow/long process. I had a similar problem with starting a large number of downloads from the context menu but I refactored and optimized that use case. That one caused a spin. I was also testing with all 600+ downloads 🤓

insidegui:
There's one more thing (ha): the last item on the list is drawing a separator, which looks weird:

image

insidegui:
> Every time a download gets canceled the table reloads. If you have a lot of downloads in the table it makes for a very slow/long process. I had a similar problem with starting a large number of downloads from the context menu but I refactored and optimized that use case. That one caused a spin. I was also testing with all 600+ downloads 🤓

Couldn't we solve this by using the runloop trick? (`NSObject.cancelPending...` and `performSelector`)

Or we could use some rx shenanigans.

insidegui:
I'm not sure if this is related to the PR or an unrelated issue, but after a couple of downloads finished, right clicking them on the sidebar was still showing the "download" and "cancel download" menu items and after choosing to delete the download using the trash button in the detail view, the file was not deleted and the UI did not reflect the change. Doing it a second time deleted the file, but the UI still did not reflect the change.

insidegui:
This is looking great!

Would it be hard to make the window/popover resizable?

allenhumphreys:
> Would it be hard to make the window/popover resizable?
Unfortunately, from what I can tell, it is not that easy.

allenhumphreys:
> I'm not sure if this is related to the PR or an unrelated issue, but after a couple of downloads finished, right clicking them on the sidebar was still showing the "download" and "cancel download" menu items and after choosing to delete the download using the trash button in the detail view, the file was not deleted and the UI did not reflect the change. Doing it a second time deleted the file, but the UI still did not reflect the change.

I believe this is because subfolders are getting created but folder monitors aren't created dynamically when downloads finish. This bug exists in `master`, I'll fix it.

allenhumphreys:
> 3 - Should clicking the progress indicator in a session's detail view still cancel the download? I believe now that we have a proper download management UI, it should either bring that up or just pause the download.

I do believe that that UI could now be reconsidered, but I'd rather not do it as part of this PR :-)

allenhumphreys:
> I believe this is because subfolders are getting created but folder monitors aren't created dynamically when downloads finish. This bug exists in master, I'll fix it.

The bug does exist in master, but my initial guess about it was wrong. Currently `existingVideoFiles` is only updated when the app is launched, which means all files that are downloaded while the app is open will behave funny.

allenhumphreys:
I tightened the layout up a little bit, using Safari as the reference. 

Before:

screen shot 2018-10-22 at 7 15 35 pm

After:

screen shot 2018-10-22 at 7 14 42 pm

insidegui:
Something about the "play" button bothers me 🤔 Maybe it looks too large compared with the "x" button...

allenhumphreys:
I know what you mean. It's definitely all in our heads though. I think it's a combination of that triangle filling the corners of its space and the center of mass being shifted to the left combined with equal spacing make it seem lopsided.

Here I've tweaked it a bit:

screen shot 2018-10-23 at 7 39 53 am

insidegui:
That looks better!

insidegui:
Tried it out on both HS and Mojave, everything checks out. Amazing work!

AnalyzePlatypus:
Looking forward to using this!

PR #514, opened 2018-10-23
Allow selecting the session title in the summary

allenhumphreys:
It bothered me a while back that I couldn't do this. Seems like something people might want to do occasionally. For example,

Me: I just watched a great WWDC session!
Friend: What was it called?
Me: ....
Me: ........
Me: ....
Me: Designing for Apple Watch Series 4
Me: I tried to copy and paste but I couldn't!

selecttitle


insidegui:
Nice quality of life update! I've been meaning to do this for ages 😅

PR #519, opened 2018-11-10
Bump deployment target to 10.13

bcmn:
Release 6.1 crashes when run on macOS 10.12 Sierra. This PR bumps the deployment target for future releases to High Sierra.


insidegui:
@bcmn Have you been able to look into it?

Related: #520

insidegui:
I have pushed some changes updating each target and the app case to require 10.13

PR #526, opened 2019-03-18
Swift 5

allenhumphreys:
Still need to handle:
non-frozen objective-c enums via `@unknown default`
Address warnings about old-style `Hashable` conformance


Cartfile
```diff
@@ -2,7 +2,7 @@ github "SwiftyJSON/SwiftyJSON" ~> 4.0
 
 github "ReactiveX/RxSwift" ~> 4.0
 
-github "realm/realm-cocoa" ~> 3.7.3
+github "realm/realm-cocoa" "tg/xcode-10.2"
```
allenhumphreys: This makes Realm work in the mean time

insidegui:
Nice!

allenhumphreys:
@insidegui @bcmn The tests are failing because of this:
https://github.com/insidegui/WWDC/blob/master/ConfCore/LiveVideosAdapter.swift#L36-L44

but the build seems ok.

WWDC/AppCoordinator+SessionActions.swift
```diff
@@ -91,6 +91,11 @@ extension AppCoordinator: SessionActionsViewControllerDelegate {
             }
         case .authorized:
             self.saveCalendarEvent(viewModel: viewModel, eventStore: eventStore)
+        @unknown default:
+            assertionFailure("An unexpected case was discovered on an non-frozen obj-c enum")
```
allenhumphreys: Assertion failures so that the app doesn't crash, but instead takes a fail silently approach in production while stopping the debugger for development. Seems unnecessary to do the `fatalError()` approach for these enums, and I really would expect more enums to be frozen going forward.

insidegui: I agree

allenhumphreys:
This should be good to merge (pending approval) as soon as https://github.com/realm/realm-cocoa/pull/6076 is merged and released which I'd expect sometime after 10.2 GM next week!

allenhumphreys:
This is now nearly completely ready, just need to actually run it, dealing with hotel internet is the worst :-(

insidegui:
Cool, let me know when it's ready

PR #530, opened 2019-03-24
Implement codable

allenhumphreys:
Updating Otávio's very old branch and addressing the feedback from the original PR #372 

Compared to #372 this narrows the scope of the conditional decoding that happens in certain places. This "conditional decoding" is where we want to explicitly allow failures. This seems to have been needed only within certain collections. This need is formalized through `ConditionallyDecodableCollection` and the `ConditionallyDecodableProtocol`. This eliminates the need for the catch-all `try?` statements and forces types to throw `ConditionallyDecodableError` to have their decoding error treated as OK.


allenhumphreys:
Pretty sure this is ready now, just in time to massively conflict with Swift 5 :-)

PR #534, opened 2019-05-02
Refine errors presented to the user

allenhumphreys:


insidegui:
Re-merging against master since the previous one was pointed at the Swift 5 branch.

WWDC/AppCoordinator.swift
```diff
@@ -316,15 +316,15 @@ final class AppCoordinator {
 
         _ = NotificationCenter.default.addObserver(forName: .SyncEngineDidSyncSessionsAndSchedule, object: nil, queue: .main) { note in
             if let error = note.object as? Error {
-                NSApp.presentError(error)
+                WWDCAlert.show(with: error)
```
sindresorhus: I would recommend leaving these as `NSApp.presentError` and instead just intercept it in the AppDelegate and use `WWDCAlert.show` there. Example: https://github.com/sindresorhus/gifski-app/blob/df25de33b43e69176a4aa481a43c70c4cfddf881/Gifski/AppDelegate.swift#L62-L65

allenhumphreys: @sindresorhus That's a great suggestion! I would've loved to have investigated that further, but mostly I changed these 2 so that this API is used exclusively throughout the app for consistency. I think the `WWDCAlert` exists for overriding the appearance on macOS < Mojave. If I weren't so busy right now I would've played further.

PR #541, opened 2019-05-14
Fixed archiving, release 6.1.2

insidegui:
Fixed archiving of builds by using Xcode's standard "copy frameworks" build phase instead of Carthage's own script (suggested by Carthage documentation).

This also releases version 6.1.2 with the latest changes and fixes.


allenhumphreys:
I'm not sure what your dSYM gathering process is like, but the `carthage copy-frameworks` command will also ensure that the dependency's dSYMs end up in the archive while the Xcode Embed Frameworks phase will only pick up dSYMs from the targets in the project. May or may not be an issue for you /shrug

allenhumphreys:
Oh I just went and read the docs, apparently they have a section for dSYMs too!

> Additionally, you'll need to copy debug symbols for debugging and crash reporting on OS X.
> 
> 1. On your application target’s Build Phases settings tab, click the + icon and choose New Copy Files Phase.
> 1. Click the Destination drop-down menu and select Products Directory.
> 1. For each framework you’re using, drag and drop its corresponding dSYM file.

insidegui:
Not gonna be an issue because I'm about to remove Fabric/Crashlytics from the project anyway 😅

PR #543, opened 2019-05-22
Fix cancelling change of downloads directory

piotr-tobolski:
Cancelling change of downloads directory acted the same way as confirming it.


insidegui:
Thanks!

PR #548, opened 2019-05-31
Keep section headers even while a filter is active

### LINKED ISSUE #226: Applying filters hides session date/time information
For those of us not at WWDC, seeing labs in the list is not useful, so the option to filter them out is great. However, filtering to just sessions hides all session date/time information so you can't tell when a session is going to happen. The nice section headers that group similar session times also disappear.

### LINKED ISSUE #430: Add section titles on filtered list
It would be nice to have the section titles showing also when the list is filtered in some way.

For example, from the schedule, it’s nice to see the videos in the order they are presented at WWDC, but I’m not really interested in the labs schedule. But if I filter the results, I really have to be looking at the description, to see which videos might be available right now.

Of course I could use the videos tab, but since those are ordered alphabetically it gets really hard to scan for new videos.
section-title

allenhumphreys:
There could be performance improvements done to this in the future, but I think it does what we want it to do

fixes #226 fixes #430

### DIFF
--- a/WWDC/SessionRowProvider.swift
+++ b/WWDC/SessionRowProvider.swift
@@ -11,25 +11,40 @@ import RealmSwift
 
 protocol SessionRowProvider {
     func sessionRowIdentifierForToday() -> SessionIdentifiable?
-    func sessionRows() -> [SessionRow]
+    func filteredRows(onlyIncludingRowsFor: Results<Session>) -> [SessionRow]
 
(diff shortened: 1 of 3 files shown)


Urkman:
please merge this...

insidegui:
We’re not merging any major changes until after WWDC week. Build from source if you’d like to use this right now. 

PR #549, opened 2019-06-04
Hide controls when the mouse exists or app resigns main

### LINKED ISSUE #538: Play-control overlay won't go away when cursor moved to a different display
### Before opening your issue, please make sure you answer YES to all of the questions below.

- [x] Are you running the latest version of the app?
- [x] Are you running a release, stable version of macOS?
- [x] Have you looked at previous issues to see if you're having an issue that's already being worked on?
- [x] If your issue is about compilation or project setup, have you followed the instructions from the readme?
- [x] Have you read the Code of Conduct, and will you uphold it's values?

- Version of macOS:
    - **10.14.4**
- Version of the app: 
    - **6.1**
- What were you trying to do when the issue happened: 
    - **Watching a session in fullscreen mode on my external display**
- Any particular hardware/software configuration you have that might be affecting the app:
    - **External display connected**
- If it is a wrong behavior, list all steps required to reproduce it:
    1. Start playing a video in full screen in external display
    2. Move the cursor back to built-in display, the overlay disappears now as expected.
    3. Click anywhere in the built-in display, the overlay appears unexpected. And it won't go away no matter it is active or idling.

yigitcanyurtsever:
Fixes two issues with controls staying on screen: 
- When using an external display and watching a full screen stream/video, clicking any other app on the other screen would show the controls. Described in #538 
- When the window is not main, you can hover over the window and the controls will show up (as expected) but won't be hidden when you leave the window. 

Fixes #538

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -1241,7 +1241,7 @@ public final class PUIPlayerView: NSView {
         guard isPlaying else { return false }
         guard player.status == .readyToPlay else { return false }
         guard let window = window else { return false }
-        guard NSApp.isActive && window.isOnActiveSpace && window.isVisible else { return false }
+        guard window.isOnActiveSpace && window.isVisible else { return false }
 
         guard !timelineView.isEditingAnnotation else { return false }
 
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -1354,7 +1354,6 @@ public final class PUIPlayerView: NSView {
         // resigning main in full screen means we're leaving the space
         if windowIsInFullScreen {
             resetMouseIdleTimer(start: false)
-            showControls(animated: false)
         }
     }
 


allenhumphreys:
This mouse hiding stuff is tricky. Well, multi-window, multi-space, multi-monitor app development is tricky. For context, it looks like this bug was introduced last year while fixing a different bug in these 2 PRs:
https://github.com/insidegui/WWDC/pull/437
https://github.com/insidegui/WWDC/pull/477

The bug in question was under a single monitor full screen playback scenario. If you switched from the full screen space, the cursor would incorrectly be hidden. So far I haven't had that issue but I'll test some more tomorrow.

The logic around this stuff is bonkers. I think it'll be hard to say for certain if this fix doesn't break a different usage scenario.

yigitcanyurtsever:
I definitely see your point. I feel like the problem comes from displaying the controls and making the cursor visible in the same method. There are some cases (like this one 😅) where it feels like the expected behavior should be to have the cursor visible but have the controls hidden. 

The app should make sure the cursor is visible every time if it the player view's window resigns main regardless of any guards or checks, i.e.`canHideControls` checks. Adding `NSCursor.unhide()` on `NSWindow.didResignMain` notification handler instead of calling `PUIPlayerView.showControls(animated:)` would fix the problem but I still think that managing the `NSCursor.hide()` and `NSCursor.unhide()` calls manually will always bring new bugs. 😅 

My suggestion here is to use [`NSCursor.setHiddenUntilMouseMoves(true)`](https://developer.apple.com/documentation/appkit/nscursor/1534665-sethiddenuntilmousemoves) on [`PUIPlayerView..mouseIdleTimerAction(_:)`](https://github.com/insidegui/WWDC/blob/master/PlayerUI/Views/PUIPlayerView.swift#L1272) instead of `NSCursor.hide()` and get rid of `NSCursor.unhide()` on `PUIPlayerView.showControls(animated:)`.  It kind of solves all cursor related issues since the cursor will be visible when the mouse moves. 

Wydt @allenhumphreys?

allenhumphreys:
@yigitcanyurtsever I think I had known about that method last year I would've probably investigated using it! But instead of getting rid of `NSCursor.unhide()` I would think we'd do `NSCursor.setHiddenUntilMouseMoves(false)`. I'd also be worried about causing situations where the controls hide but the curse is somehow still visible in full screen, single monitor viewing, but I might be over-thinking it.

So personally I'm open to these suggestions but would love a cursory review from @insidegui !

insidegui:
During my testing, I didn't find any issues with this PR applied. We'll see if I regret this decision later 😅

PR #558, opened 2019-06-06
Disables the global alert when network errors are encountered

allenhumphreys:


mackuba:
Oh man, thank you 🙏 I'm getting this many times a day because my WiFi is somehow messed up at the moment, it's driving me crazy...

PR #570, opened 2019-06-11
Add preference to automatically delete videos when watched

### LINKED ISSUE #567: Auto-delete watched videos
It would be cool to provide an option in the settings to auto-delete watched session videos so they do not clutter up disk space.

# Specs
- Version of macOS: 10.14.5
- Version of the app: 6.1.2

bcmn:
Will fix #567 

I just need to wrestle with Interface Builder to add the toggle itself 🙃


WWDC/PlaybackViewModel.swift
```diff
@@ -126,6 +126,17 @@ final class PlaybackViewModel {
 
                 if !d.isZero { self.nowPlayingInfo.value?.progress = p / d }
             }
+
+            NotificationCenter.default.addObserver(self, selector: #selector(self.playerDidReachEndOfVideo(note:)), name: NSNotification.Name.AVPlayerItemDidPlayToEndTime, object: nil)
```
allenhumphreys: This notification only fires if they play to the very, _very_ end, which won't necessarily align with the other notions we have of "watched". We also probably shouldn't delete the video that is currently being played. Are there any other integration points for a feature like this?

insidegui: Yeah, it should be deleted when the app considers it as being watched. And there's also the consideration of when to actually delete the video, since deleting the video file that's currently being played sounds dangerous.

insidegui: Another question is: should we delete the video if the user manually marks it as watched?

insidegui: Now I'm starting to wonder if this feature is even worth it. Is it really that hard to press a button after you finish watching the video? Maybe we could display the delete option more prominently over the video when it's near the end? 🤔

allenhumphreys: Post roll overlay? Like on Net... uh, major video subscription services ;-)

allenhumphreys: If we went a fully automatic route I'd be in favor of a cleanup job on app launch. Also, just confirmed users can delete the currently playing video manually already 😂😭

insidegui: What happens if you do? 😛

allenhumphreys: Everything works fine. I'm guessing due to having a robust OS and frameworks that don't let the file get completely removed when it's open? Or maybe AVFoundation defensive programming?

bcmn:
Closing, for all the reasons listed above 😛 We talked about this a little more elsewhere, & it's not immediately obvious how to tie it in with the existing definition of "Watched" without more work than I think we're willing to give at the moment, & when's best to safely delete the files.

PR #581, opened 2019-09-12
Add JKL playback controls shortcuts

hvsw:
# Introduction
JKL playback controls shortcuts are pretty common in video-related software like Quicktime, Final Cut Pro X, YouTube, etc.

Currently, the behavior for QuickTime and Final Cut Pro are very similar to this:
[](https://support.apple.com/kb/PH12779)

As a video player, and not a video editing software, YouTube has a different, and simpler, approach:
Screen Shot 2019-09-12 at 16 40 38

I use this so often that when I tapped the keys and this didn't work out of the box I thought my 2018 MacBook Pro keyboard had more failing keys lol

Last but not least, I love these shortcuts so I added this, maybe someone else can love it too.

# Implementation
Since the WWDC app is a video player, and not a video editing software, I'd say the YouTube's approach makes more sense here, so I propose to add the following shortcuts to the player:
`J` - Go back N seconds
`K` - Play/Pause
`L` - Go forward N seconds

> `N` is the number of seconds configured by the user, which currently the possible values are 15 or 30

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -1171,15 +1171,15 @@ public final class PUIPlayerView: NSView {
             guard !self.timelineView.isEditingAnnotation else { return event }
 
             switch command {
-            case .spaceBar:
+            case .spaceBar, .k:
                 self.togglePlaying(nil)
                 return nil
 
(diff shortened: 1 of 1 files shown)


insidegui:
Thank you for this incredibly well-made PR 😊

PR #585, opened 2019-10-01
Fix Ctrl-Click on session list row

subdan:
I removed this line of code because it breaks Ctrl+Click action that calls context menu.
Also this fixes issue with bad row selection color. Row selection color is very light. Text color merges with background color.


insidegui:
Resolved in #588, since this solution caused some other issues. Thank you for your contribution!

PR #588, opened 2019-12-01
Revert cell selection drawing behavior and restore ctrl+click

allenhumphreys:
• Reverts selection drawing to the way it used to be
• Restores ctrl+click functionality that I broke
• Also for some ✨ , I added a conditional for when the window is not key/app is not active for the selection color to go muted like the default NSTableView behavior

Basically, I broke this behavior a while ago while attempting to track down a drawing performance issue. The changes I made were not actually the fix, but I never reverted them.

Supersedes #585 

NewSelection


allenhumphreys:
This change appears to affect the background color of the section headers, but I think it actually fixes it as compared to the original. I think my code may've broken that too.

PR #590, opened 2020-02-21
Fix popover positioning on Catalina

allenhumphreys:
Screen Shot 2020-02-21 at 9 51 33 AM
Screen Shot 2020-02-21 at 9 51 10 AM

I hope this fix will work on older macOS versions, too. I'm not sure why I had that height constraint in there to begin with.


insidegui:
Merged manually. Tested in Mojave and it was all good.

PR #592, opened 2020-04-05
Fixes 2020 feed issues

insidegui:
This release fixes an issue that occurred when we pointed the API proxy to Apple's 2020 endpoint.

The problem was that it includes the 2020 edition with placeholder dates and `current = true`, but there are no session instances, so our initialization dependencies filter, which wanted instances to not be empty, was causing the app to appear blank.

There's still an issue where the Schedule tab will appear empty with no useful message to the user as to why, that should be fixed in a future update. The focus of this release is to finally release many things that have been implemented and also be able to point to the new feed (which added some new content).


WWDC/AppCoordinator.swift
```diff
@@ -257,15 +257,15 @@ final class AppCoordinator {
         // Initial app launch waits for all of these things to be loaded before dismissing the primary loading spinner
         // It may, however, delay the presentation of content on tabs that already have everything they need
 
-        let starupDependencies = Observable.combineLatest(storage.tracksObservable,
+        let startupDependencies = Observable.combineLatest(storage.tracksObservable,
```
allenhumphreys, line 260: Pretty sure this typo was mine 🙈. I do this at work all the time, too!

insidegui: I have a colleague that does it all the time, so I'm used to fixing typos in code review :P

PR #595, opened 2020-05-25
Implemented Apple's transcripts API

insidegui:
This implements support for Apple's own transcripts API, which was added to their app a while ago. Transcripts are provided in a few languages, so I made language switching possible with a new preference.

The first time a version of the app with the new transcripts infrastructure is run, it will re-download all transcripts again from Apple's service. The entire process takes less than a minute on my network.


bcmn:
I managed to get this to crash with a `BAD_INSTRUCTION` exception on `SyncEngine.swift#L153` -- the scenario I was in superficially seemed to be "changing the preference while the Transcription tab was open", but I'm not confident that this means anything. I have some logs I can pass on -- it seems the transcript downloading & indexing had completed, the last relevant log seems to be:
`2020-05-27 18:03:33.758972+0100 TranscriptIndexingService[18028:738023] [TranscriptDownloader] Handing out 671 transcript(s) for storage`
which made me wonder if saving to database hadn't completed when it failed?

insidegui:
@bcmn That was caused by calling `resume()` in the XPC connection out of place. I took the opportunity to modernize the whole XPC implementation, moving everything to ConfCore and just starting the service in the TranscriptIndexingService (which has always linked against ConfCore anyways).

The next step is to lock the language pref while indexing is in progress. I explored the idea of allowing the process to be cancelled in flight if the language pref was changed, but decided it was too much of a hassle for something that only takes around 30s and won't be changed often.

Wait for the next commit before merging this :)

insidegui:
Done! It will now disable the popup menu while indexing is in progress and show the progress in place.

Screen Shot 2020-05-27 at 16 53 04

Transcripts/Source/Networking/Loader.swift
```diff
@@ -0,0 +1,35 @@
+//
+//  Loader.swift
+//  Transcripts
+//
+//  Created by Guilherme Rambo on 25/05/20.
+//  Copyright © 2020 Guilherme Rambo. All rights reserved.
+//
+
+import Foundation
+import Combine
```
bcmn, line 10: I don't think this is used? I was very intrigued for a moment haha.

Issue #580, opened 2019-07-24
Database compression


insidegui:
As described in #578, the ConfCore.realm file can get quite big. Realm offers a way to re-save a database file in a way that compresses it. I remember trying to do this automatically a long time ago but having trouble getting it to work reliably.

Maybe something changed that would allow us to implement this now, or at least offer a manual way for the user to do so that doesn't require downloading another piece of software.

bcmn:
There's some intriguing [Realm docs about larger-than-expected Realm files](https://realm.io/docs/swift/latest/#limitations-file-size), especially around accessing while using GCD. I've not had a chance to work out if any of these might be describing what's happening here.

I've seen some fairly "nuclear option" implementations of forcing a re-compression, where people will write another realm file and swap it out. 😶

insidegui:
I believe the `shouldCompactOnLaunch` thing is what I tried before and it caused an awful hang when launching the app. We could try it again to see if it's better now tho.

bcmn:
🤔 Interesting, okay.

mackuba:
@insidegui I've just been comparing some Time Machine snapshots to see why it sometimes needs to back up more than I would expect, and I found an interesting file in the diff - a `~/Library/Application Support/io.wwdc.app/ConfCore.realm` weighing 38.65 GB… this is not normal, is it? 😬

insidegui:
We've had some builds in the past that entered sync loops, causing the database file to grow in that way. If you've been using the app for a long time, it's possible you were affected. Unfortunately, there's no easy solution I can recommend besides just nuking your database file and starting over :/

mackuba:
What do I lose if I do that?

insidegui:
Favorites, bookmarks and video progress (watched/unwatched). The rest is downloaded from the service.

mackuba:
So favorites and watched status aren't synced between computers?

insidegui:
They are. If you’re using iCloud sync, they should be restored. Keep a copy of the original db file just in case.

mackuba:
Ok, thanks 👍

insidegui:
Handled in #596 

Let us know how it goes for your database @mackuba. I expect it may hang the first time you launch the app with this enabled, but afterwards it should be fine.

insidegui:
Just FYI: this has been temporarily disabled in version 7.1 as we wait for Realm to fix https://github.com/realm/realm-cocoa/issues/6652

PR #596, opened 2020-05-26
Enabled database compression on launch

insidegui:
This aims to solve what's described in #580 

I haven't tested with a really huge database file, but with a reasonably large database file (~300MB), the compression occurred at launch with no noticeable impact to the user experience.

A defaults flag was introduced (`WWDCDisableDatabaseCompression`) to disable the feature, just in case.

### DIFF
--- a/ConfCore/Storage.swift
+++ b/ConfCore/Storage.swift
@@ -19,11 +19,27 @@ public final class Storage {
     public let realm: Realm
 
     let disposeBag = DisposeBag()
-    private let log = OSLog(subsystem: "ConfCore", category: "Storage")
+    private static let log = OSLog(subsystem: "ConfCore", category: "Storage")
+    private let log = Storage.log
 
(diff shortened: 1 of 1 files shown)


mackuba:
Should I try it on my 38-gig file (I haven't reset it yet)? Should it help in this case?

insidegui:
@mackuba That would be great! Thank you

mackuba:
Looks like it worked! I tried it last night, and at first it said the database was a few hundred MB, but I've realized it was running in debug mode and pointing to a different directory, then I rerun it in release mode and it just said `Total bytes: 0`… and now I've just launched it again and it seems it has compacted it to a few hundred MB:

Before:

Screen Shot 2020-05-30 at 01 50 06

After:

Screen Shot 2020-05-30 at 19 11 27

insidegui:
Awesome! Thanks for taking the time to test this @mackuba

PR #597, opened 2020-05-27
Limit downloads folder recursion, add warnings (Closes #584)

### LINKED ISSUE #584: If user sets directory with a lot of files as Download folder, app will freeze
- Version of macOS
10.14.6 (18G103)
- Version of the app
6.1.3
- What were you trying to do when the issue happened
I changed Download folder to ~/Documents
- If it is a hang (beach ball), please provide a process sample
I used Instruments.app and discovered that
`DownloadManager.updateDownloadedFlagsByEnumeratingFilesAtPath`
freezes the app because I changed Downloads folder to ~/Documents where I store all my files (50 Gb).
```
guard let enumerator = FileManager.default.enumerator(atPath: path) else { return }
guard let files = enumerator.allObjects as? [String] else { return }

```
`enumerator.allObjects` freezes the app because I have a lot of files in the Documents directory.

Screenshot 2019-09-29 at 20 48 02

As a solution, we can disallow choosing an non-empty directory as Download folder.

Also there is no button to reset Download folder to default value. I think we should add this button. I can open Pull request with this button.

insidegui:
This addresses the issue that caused hangs when a downloads folder had a deeply nested hierarchy of files.

I've implemented both the suggestion by @allenhumphreys to limit recursion, but I've also added some warnings to the UI when changing folders.

If the user tries to set the downloads folder to a folder that has more than 3 files inside of it (3 was randomly chosen as a threshold for "is this folder empty enough?"), there will be a warning:
Screen Shot 2020-05-27 at 18 02 39

If the user tries to do something *really stupid*, such as using the home directory or `/` as the downloads folder, the app will prevent it:
Screen Shot 2020-05-27 at 18 06 58

### DIFF
--- a/WWDC/GeneralPreferencesViewController.swift
+++ b/WWDC/GeneralPreferencesViewController.swift
@@ -162,7 +162,7 @@ class GeneralPreferencesViewController: NSViewController {
         downloadsFolderLabel.stringValue = url.path
     }
 
-    private lazy var dangeoursLocalStoragePaths: [String] = {
+    private lazy var dangerousLocalStoragePaths: [String] = {
         [
             NSHomeDirectory(),
             "/"
@@ -179,7 +179,7 @@ class GeneralPreferencesViewController: NSViewController {
     }
 
     private func validateDownloadsFolder(_ url: URL) -> Bool {
-        guard !dangeoursLocalStoragePaths.contains(url.path) else {
+        guard !dangerousLocalStoragePaths(url.path) else {
             showStoragePathError(with: "This folder can't be used to store downloaded videos, please choose another one. Downloaded videos can't be stored in the root of your home directory or in the root of the filesystem.")
             return false
         }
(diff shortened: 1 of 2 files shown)


WWDC/GeneralPreferencesViewController.swift
```diff
@@ -143,12 +143,71 @@ class GeneralPreferencesViewController: NSViewController {
...
+        let alert = NSAlert()
+        alert.messageText = "Folder can't be used"
+        alert.informativeText = message
+        alert.alertStyle = .critical
+        alert.addButton(withTitle: "OK")
+        alert.runModal()
+    }
+
+    private func validateDownloadsFolder(_ url: URL) -> Bool {
+        guard !dangeoursLocalStoragePaths.contains(url.path) else {
```
bcmn, line 182:
```suggestion
    private lazy var dangerousLocalStoragePaths: [String] = {
        [
            NSHomeDirectory(),
            "/"
        ]
    }()

    private func showStoragePathError(with message: String) {
        let alert = NSAlert()
        alert.messageText = "Folder can't be used"
        alert.informativeText = message
        alert.alertStyle = .critical
        alert.addButton(withTitle: "OK")
        alert.runModal()
    }

    private func validateDownloadsFolder(_ url: URL) -> Bool {
        guard !dangerousLocalStoragePaths(url.path) else {
```

PR #599, opened 2020-05-28
Showing "coming soon" info when the schedule is not available

insidegui:
I'm not sure if this is something that's going to happen every year going forward, but this year the 2020 edition was marked as "current" before even the placeholder schedule was available.

In its place, Apple is using an "article" that displays an image and some text about the upcoming event. I chose to parse their information on the backend and return it together with the config response we were already using for the transcript feeds (we should probably start looking at the config for all URLs in the future).

Here's what the Schedule tab looks like when there's no schedule available and an event "article" in the API:
Screen Shot 2020-05-28 at 12 20 37

And here it is when the schedule is not available and there's no event "article" in the API (it just shows a generic "check back later" message):
Screen Shot 2020-05-28 at 12 20 04


ConfCoreTests/DecodingTests.swift
```diff
@@ -323,7 +323,7 @@ class AdapterTests: XCTestCase {
     }
 
     func testConfigAdapter() {
-        let config = loadFixture(from: "config", type: RootConfig.self)
```
bcmn, line 326: It's probably worth adding an "eventHero" key to the fixture data & testing the decoding of the object, for how little effort it takes?

insidegui:
Hmmm the transcript indexing process is never starting for me when running from this branch, works fine on master. I'll investigate 🧐

PR #600, opened 2020-05-29
Bumped deployment target, modernized design

insidegui:
This raises the deployment target to 10.15, as discussed in the private Slack group.

I also took the opportunity to modernize the look a little bit, adopting the SF Rounded font for large titles, adjusting fonts and colors in several places, and changing the main session action icons to assets based on SF Symbols, to match the title font a little better.

This also introduces something I copied from Xcode. We've had problems with highlights lacking contrast when the user sets their system accent color to graphite. To fix that, I'm using a faded blue color as the accent color when that's the case. Several places in the app aren't updating their accent colors in real time when that's changed in System Preferences, that should be fixed in a future pull request.

Screen Shot 2020-05-29 at 20 04 54
Screen Shot 2020-05-29 at 20 04 44
Screen Shot 2020-05-29 at 20 04 40


insidegui:
Oh, I forgot to mention that this also completely removes pre-v6 migration code. Version 6 has been around long enough for that to not be necessary anymore.

PR #603, opened 2020-05-30
Skip intro feature

insidegui:
This implements a "skip intro" feature. The way it works is it looks for the time of first transcript annotation for that session and starts playback from that time. This won't work for sessions with no transcript, but since we're now using Apple's official transcripts API, the vast majority of sessions do have a transcript.

I also took the opportunity to move playback-related preferences to a separate "playback" tab, which currently has only two settings, but will likely have more in the near future 😇
Screen Shot 2020-05-30 at 17 07 14


bcmn:
Screenshot 2020-05-31 at 14 38 30

I think there's something weird going on in the iCloud syncing switch -- & we could probably collapse the spacing a bit?

insidegui:
@bcmn the reason you’re seeing it like that is because your build doesn’t have the iCloud feature. That visual issue doesn’t show up in release builds with iCloud enabled.

PR #605, opened 2020-06-02
Increase control contrast, slightly compact player controls

### LINKED ISSUE #572: Improve colour contrast on Playback progress indicator
There are three colours in the playback progress indicator -- the current time, the hovered time, & the background colour. The WCAG contrast ratio (not normally applied to two non-font elements, but 🤫) doesn't pass on simply the current + background colour, never mind the intermediary hover colour.

Screenshot 2019-06-12 at 22 26 44

I suggest we increase the contrast between these elements to increase legibility.

bcmn:
Fixes #572

This PR started as a review of the playback timeline control colours, & kind of expanded out from there -- first reintroducing the background gradient to help with contrast, then eventually replacing that in favour of a more modern solution.

**Before...**
Screenshot 2020-06-02 at 17 40 33

**After...**
Screenshot 2020-06-02 at 17 48 30

I'm a little wary of how muddy the "time preview" is now -- the accent colour timestamp which appears on mouseover. I might have another go at giving it a rounded rect background separately.


PlayerUI/Views/PUIPlayerView.swift
```diff
@@ -740,22 +737,21 @@ public final class PUIPlayerView: NSView {
...
+
         addSubview(scrimContainerView)
         addSubview(controlsContainerView)
 
         scrimContainerView.translatesAutoresizingMaskIntoConstraints = false
         scrimContainerView.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
         scrimContainerView.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
-        scrimContainerView.topAnchor.constraint(equalTo: topAnchor).isActive = true
         scrimContainerView.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
+        scrimContainerView.heightAnchor.constraint(equalTo: controlsContainerView.heightAnchor, multiplier: 1.4, constant: 0).isActive = true
```
bcmn, line 749: Not super happy with use of the multiplier here -- I'm not sure if I'm anchoring to the wrong thing, but the height wasn't behaving as expected at all 🤔

insidegui (approved):
Nice!

Issue #511, opened 2018-10-17
"Save a copy..." for Downloaded Sessions


allenhumphreys:
Just a thought I had while working on the downloads window. Sometimes I'd like to trim down a clip to send to someone and finding it/getting a copy of it to work with would be handy. Could go in the share menu possibly.

I don't think we want to encourage a "Reveal in Finder" type mechanism as users doing anything to the files other than deleting them could be disastrous.

insidegui:
Sounds like a good idea and not too hard to implement. 

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

bcmn:
I've really liked the idea of being able to bookmark a chunk of video, instead of a point, for quite a while now. I wonder if this idea would work well with it -- the ability to either link to the chunk, or export the chunk as a new file to share. 🤔

allenhumphreys:
My workflow is to get the file, trim with QuickTime and then share.

insidegui:
> My workflow is to get the file, trim with QuickTime and then share.

That's not too hard to implement with `AVFoundation`

allenhumphreys:
No, but the UI though... big can of worms turning the app into an editor.

insidegui:
I believe `AVKit` has a built-in control that can handle trimming.

insidegui:
Just had a thought about how cool this feature could be for sharing short clips from the videos, but then I thought about copyright laws and got depressed 😑

PR #606, opened 2020-06-02
Clip sharing (Implements #511)

insidegui:
This adds the ability to export selected portions of a session's video. The main limitations are that the session must be downloaded, and I also added a time limit of 5 minutes because the feature was designed for sharing short clips and not huge portions (or entire) sessions.

By default, the app will add a little tag at the top of the exported clip with the title of the session and some other information, there's a new preference to turn that off if the user prefers.

Screen Shot 2020-06-02 at 20 38 07

The UI is pretty simple. The shelf view / player is replaced by an `AVPlayerView` in trimming mode, while the rest of the app's UI is covered by a dark mask which prevents interacting with it. 

Another little touch is that when you enter sharing mode while playing a video, the app will pause and set the starting point of the shared clip to the current playback position.

I'm sure there are ways to break this I didn't think of, but for the common use case of quickly sharing a piece of video, it has worked really well for me.

Screen Shot 2020-06-02 at 20 20 32

Screen Shot 2020-06-02 at 20 36 12


insidegui:
Fixed, also implemented some additional niceties :)

PR #627, opened 2020-07-03
Adjust how realms are initialized for some crashy paths

allenhumphreys:


ConfCore/TranscriptIndexer.swift
```diff
@@ -29,7 +29,7 @@ public final class TranscriptIndexer {
         self.manifestURL = manifestURL
     }
 
-    fileprivate let queue = DispatchQueue.global(qos: .utility)
+    fileprivate let queue = DispatchQueue(label: "Transcript Indexer", qos: .background)
```
allenhumphreys, line 32: Realm only works with serial queues.

ConfCore/TranscriptIndexer.swift
```diff
@@ -102,7 +102,7 @@ public final class TranscriptIndexer {
         )
 
         do {
-            let realm = try storage.makeRealm()
+            let realm = try storage.makeRealm(on: nil)
```
allenhumphreys, line 105: This might actually supposed to be `storage.realm` instead. At this point in the code it is on the XPC thread which is where the underlying realm was created.

ConfCore/TranscriptIndexer.swift
insidegui: True. Using `.global` is usually not a good idea anyways 😬

PR #628, opened 2020-07-04
Fix session summary string replacement, be smarter about the default selection

allenhumphreys:
…selection

### DIFF
--- a/WWDC/SessionSummaryViewController.swift
+++ b/WWDC/SessionSummaryViewController.swift
@@ -200,8 +200,9 @@ class SessionSummaryViewController: NSViewController {
 
         viewModel.rxSummary.subscribe(onNext: { [weak self] summary in
             guard let self = self else { return }
-            let range = NSRange(location: 0, length: self.summaryTextView.string.count)
-            self.summaryTextView.textStorage?.replaceCharacters(in: range, with: self.attributedSummaryString(from: summary))
+            guard let textStorage = self.summaryTextView.textStorage else { return }
+            let range = NSRange(location: 0, length: textStorage.length)
+            textStorage.replaceCharacters(in: range, with: self.attributedSummaryString(from: summary))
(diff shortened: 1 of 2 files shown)


WWDC/SessionSummaryViewController.swift
```diff
@@ -200,8 +200,9 @@ class SessionSummaryViewController: NSViewController {
 
         viewModel.rxSummary.subscribe(onNext: { [weak self] summary in
             guard let self = self else { return }
-            let range = NSRange(location: 0, length: self.summaryTextView.string.count)
```
allenhumphreys, line 203: I found that the string count isn't always the length of the text storage, resulting in cases of some mysterious characters at the end of the new summary!

WWDC/SessionsTableViewController.swift
```diff
@@ -261,12 +261,29 @@ class SessionsTableViewController: NSViewController, NSMenuItemValidation {
...
                     }
 
-                    selectedIndexes = IndexSet(newRowsSet.intersection(selectedRows).map { $0.index })
+                    let newSelection = newRowsSet.intersection(previouslySelectedRows)
+                    if let topOfPreviousSelection = previouslySelectedRows.first, newSelection.isEmpty {
+                        // The update has removed the selected row(s).
+                        // e.g. You have the unwatched filter active and then mark the selection as watched
+                        stride(from: topOfPreviousSelection.index, to: -1, by: -1).lazy.compactMap {
+                            return IndexedSessionRow(sessionRow: oldValue[$0], index: $0)
+                        }.first { (indexedRow: IndexedSessionRow) -> Bool in
```
allenhumphreys, line 275: As a result of writing Kotlin for 1.5 years, I feel lost in Swift. But I don't think you can get around all the nulls. I could stack some `if let`s probably

PR #632, opened 2020-07-13
Update filter preservation to use Codable compatible serialization to facilitate removing SwiftyJSON

allenhumphreys:


WWDC/FilterType.swift
```diff
@@ -8,47 +8,13 @@
...
-    var identifier: String { get set }
+    var identifier: FilterIdentifier { get set }
     var isEmpty: Bool { get }
     var predicate: NSPredicate? { get }
     mutating func reset()
-    func dictionaryRepresentation() -> WWDCFilterTypeDictionary
-
-}
-
-// Can't use WWDCFilterTypeDictionary as Value's type, likely a bug in Swift
```
allenhumphreys, line 24: _narrator voice_
It was a bug in Swift

PR #670, opened 2023-05-28
Remove unused Rx bits

allenhumphreys:
I'm basing the removal of RxSwift on your branch, and here are some small pieces of house keeping that are safe to take into the base branch now! (Also, master won't build for me because of some package nonsense)


bcmn (approved):

allenhumphreys:
@bcmn thanks for the approve! Wasn't sure if you were still watching. I reformulated this cleanup in #671 against master and included the SPM fixes from Gui's branch so master will be buildable sooner than later.

PR #671, opened 2023-05-28
Remove RxSwift in favor of combine

allenhumphreys:
- [x] App startup is a flickering mess
- [x] Check for performance optimizations on the streams especially with regard to realm
- [x] Refine the UI binding extensions dealing with Error == Never and the default values and such. Also, add ones to drive an @Published UI property
- [x] TouchBar/Now Playing test?
- [x] CloudKit Code


insidegui:
This is way smaller than I thought it would be 😅

We'll definitely wait until after next week's keynote before merging this, since we already have a lengthy PR (#669) for this year's update, but I'm definitely in favor of doing it.

allenhumphreys:
@insidegui  I think this should probably get merged in so you can start building on it too. I don't think it's perfect, but there's some good fixes in here.

I cannot test Touch Bar or CloudKit stuff, so there's a few places that are #if 'd away that may need to be migrated to Combine!

WWDC/AppCoordinator.swift
```diff
@@ -248,32 +251,49 @@ final class AppCoordinator: Logging {
...
+            .receive(on: DispatchQueue.main)
+            .sink { [weak self] tracks, _, _, sections in
+                guard let self else { return }
 
                 self.tabController.hideLoading()
-                self.searchCoordinator.configureFilters()
+                if !hasPerformedInitialListUpdate {
+                    // Filters only need configured once, the other stuff in
+                    // here might only need to happen once as well
+                    self.searchCoordinator.configureFilters()
```
allenhumphreys, line 281: This was causing filters to be reset after the content sync, I put a gate around it for now, but the long term solution is different than this.

insidegui: OMG thanks, this has been driving me nuts!

insidegui:
@allenhumphreys I don't have Touch Bar hardware either, but there's a simulator in Xcode. Window > Touch Bar > Show Touch Bar. Could you test with that and let me know?

I'll check out the CloudKit stuff.

WWDC/DownloadManager.swift
```diff
@@ -178,109 +177,102 @@ final class DownloadManager: NSObject, Logging {
...
+        }
 
-                let error = note.userInfo?["error"] as? Error
-                observer.onNext(.failed(error))
+        let progress = nc.publisher(for: .DownloadManagerDownloadProgressChanged, filteredBy: asset.remoteURL).map { note in
+            if let info = note.userInfo?["info"] as? DownloadInfo {
+                latestInfo = info
+                if info.taskState == .suspended {
+                    // We can get progress updates that were from while the task was suspending
+                    return DownloadStatus.paused(info)
```
allenhumphreys, line 236: @insidegui This fixes the bug where the pause/resume button gets into the wrong state which has been there since I implemented it like 4 years ago!

allenhumphreys:
> Could you test with that and let me know?

@insidegui It's still working as expected, I think the icons will get an update when you finish your branch, so remember to check the Touch Bar buttons in that branch!

WWDC/AppCoordinator.swift
allenhumphreys: The original reason it's always called is because it uses values from Storage. But the 1st setup and the update need to be separated so that they can take any user modifications into account after the first load. By blocking this, we're potentially missing an update to filtering capabilities, but that's a small thing for now!

insidegui:
I wonder if we should even keep Touch Bar support in. It’s not trivial to support and Apple has made it quite clear that it’s a dead feature.

allenhumphreys:
If the now playing stuff gives the basic controls in the touchbar... seems like that could make sense. On the other hand there are a lot of touchbar macs out there.

PR #674, opened 2023-05-30
Remove WWDCAgent and related code

insidegui:
The agent was only used by a Raycast integration that's no longer available. As a nice side-effect, this fixes some app foreground activation issues, including one where the active tab would not be properly drawn with the accent color.


allenhumphreys (approved):
All i did was confirm LSUIElement was removed from the info.plist, the rest I'll leave up to fate! (I'm currently confined to my phone)

PR #677, opened 2023-05-31
Use modern unified logging API

allenhumphreys:
- Removes all usage of `os_log` in favor of Logger.<level>
- Codifies logging via a protocol to allow globally tuning how logging is done
- PlayerUI is excluded because it does not depend on ConfCore
- `Logging` is currently in `ConfCore` but might make more sense as a small package
- The subsystem's we're using don't follow Apple's recommendation of reverse DNS
- Considering a few of the above things, we might want to tweak the subsystem derivation (could just be the main bundle ID by default, as an example)


Packages/ConfCore/ConfCore/BookmarkSyncObject.swift
```diff
@@ -50,10 +50,7 @@ extension Bookmark: SyncObjectConvertible, BelongsToSession {
             do {
                 bookmark.snapshot = try Data(contentsOf: snapshotURL)
             } catch {
-                os_log("Failed to load bookmark snapshot from CloudKit: %{public}@",
-                       log: .default,
-                       type: .fault,
-                       String(describing: error))
+                Logger.default.fault("Failed to load bookmark snapshot from CloudKit: \(String(describing: error), privacy: .public)")
```
insidegui, line 53: Should we introduce a custom `Logger` instance for these? I don't like using `Logger.default` these days because it doesn't really help much when you're trying to find a specific message in the logs.

allenhumphreys: We can. I was trying to stick with the old rule of "during a refactor, don't fix things unless you want to break things". But it's probababably ok in this case (famous last words, lol)

allenhumphreys: I'm taking a look at how this all might go a little further.

insidegui (approved):

PR #679, opened 2023-06-01
Implemented data reset option

insidegui:
When holding the Option key during app launch, it will offer to reset the local database. Confirming the reset deletes the application support directory with the local database and caches. This is an escape hatch in case existing local data is causing the app to misbehave.

https://github.com/insidegui/WWDC/assets/67184/5e96224e-59d7-4d5c-83d6-a75307da4509


allenhumphreys (approved):
Great feature implementation!

PR #693, opened 2023-06-07
Implement keyboard localization support for player key commands

boettges:
The changes enable consistent keyboard shortcuts for media player commands across different keyboard layouts irregardless of the localization. This is achieved by identifying commands based on the character output of keys instead of their key codes which may vary. Due to the non-textual nature of the arrow keys their input remains to be identified by their keycodes.

Example:
The current mapping for `minus` and `plus` is set to the keycodes `27` and `24`. 
This is not applicable to all keyboard layouts. For example, German localized ones use `44` and `30` respectively. 
The proposed changes should solve this. 

Thank you in advanced for reviewing and thank you for the handy app 👍

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -1140,15 +1140,37 @@ public final class PUIPlayerView: NSView {
 
     private var keyDownEventMonitor: Any?
 
-    private enum KeyCommands: UInt16 {
-        case spaceBar = 49
-        case leftArrow = 123
-        case rightArrow = 124
-        case minus = 27
-        case plus = 24
-        case j = 38
-        case k = 40
-        case l = 37
+    private enum KeyCommands {
+        case spaceBar
+        case leftArrow
+        case rightArrow
+        case minus
+        case plus
+        case j
+        case k
+        case l
+
+        static func fromEvent(_ event: NSEvent) -> KeyCommands? {
+            switch event.keyCode {
+            case 123: return .leftArrow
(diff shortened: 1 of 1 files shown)


PlayerUI/Views/PUIPlayerView.swift
```diff
@@ -1140,15 +1140,37 @@ public final class PUIPlayerView: NSView {
...
+        case l
+
+        static func fromEvent(_ event: NSEvent) -> KeyCommands? {
+            switch event.keyCode {
+            case 123: return .leftArrow
+            case 124: return .rightArrow
+            default: break
+            }
+            
+            guard let character = event.charactersIgnoringModifiers else {
```
allenhumphreys, line 1160: Reading through the docs, `keyCode` and `charactersIgnoringModifiers` both say they they can raise exceptions if the event is not a key event.

`addLocalMonitorForEvents(matching: .keyDown)` should ensure it's always a key event, but this function could use a `guard` for the type of event for thoroughness.

To increase the accuracy of this function, we could also ensure no modifiers are pressed as well via `event.modifierFlags.isEmpty`.

allenhumphreys (approved):
Overall this is a great improvement! If you have time could you consider the adjustments I suggested?

If you don't have the time, please let us know, we could probably take this as is as both of the critical suggestions I made also apply to the existing code!

insidegui:
Thanks @boettges!

@allenhumphreys I'll apply your suggestions and submit another PR just with them.

boettges:
@allenhumphreys Those are very valid suggestions. Thank you both!

PR #694, opened 2023-06-07
Adjust transcripting skipping logic

allenhumphreys:
Remove year based predicate when determining whether transcription should be attempted.

We should consider potentially remove all skipping logic accept etag based skipping on the realm updates.

### DIFF
--- a/WWDC/TextualFilter.swift
+++ b/WWDC/TextualFilter.swift
@@ -28,7 +28,7 @@ struct TextualFilter: FilterType {
 
     var predicate: NSPredicate? {
         guard let value = value else { return nil }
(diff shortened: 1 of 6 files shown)


WWDC/TextualFilter.swift
```diff
@@ -28,7 +28,7 @@ struct TextualFilter: FilterType {
 
     var predicate: NSPredicate? {
         guard let value = value else { return nil }
-        guard value.count > 2 else { return nil }
```
allenhumphreys, line 31: Sneaking this in because you can't search for vr/ar currently.

insidegui (approved):

Issue #657, opened 2021-06-08
Subtitle button is missing after download video


young40:
- Version of macOS: 10.15.7
- Version of the app: 7.3.3
- What were you trying to do when the issue happened
After download video
![Screen Shot 2021-06-08 at 8 51 29 PM](https://user-images.githubusercontent.com/1680953/121197654-0fc9b080-c8a4-11eb-96e9-bdb09c3add19.png)

insidegui:
This has been a long-standing issue (as seen in #286). Basically the downloaded videos don't include the subtitles (at least not in a way that our player, which is just a wrapper around AVPlayer, understands). So if you rely on subtitles, you'll have to use streaming for now. I'll leave this issue open so that others can reference it.

sugarmo:
> This has been a long-standing issue (as seen in #286). Basically the downloaded videos don't include the subtitles (at least not in a way that our player, which is just a wrapper around AVPlayer, understands). So if you rely on subtitles, you'll have to use streaming for now. I'll leave this issue open so that others can reference it.

Is it possible to convert transcript to subtitles? Putting text in front of the video player should not be that hard (just my opinion).

insidegui:
Tracking this in #267

PR #695, opened 2023-06-08
Player UI Modernization

insidegui:
- [x] Use SF Symbols for most controls
- [x] Add AirPlay button
- [x] Make player controls more compact
- [x] Player controls positioned/sized according to video content instead of whole player area

### Before

PUIBefore

### After

CleanShot 2024-05-29 at 21 26 27


allenhumphreys:
Looks great, noticed a few things:
1. The window width is now constrained to be unnecessarily wide
2. The book marks indicators a bit funny, could be existing issues:
Screenshot 2023-06-08 at 1 50 20 PM
I managed to create a bunch of book marks all at the same spot, then while removing them I managed to get the app to crash with a constraint not being finite! You can also crash it by just changing the playback speed a few times.
3. This isn't from your work, but I noticed full screen mode doesn't use the external player status which feels inconsistent. In full screen, you just see the video poster image. This is likely because the entire player view moves to the full screen window.
4. For the PiP external status view, the status view takes the full space of the view which results in a bit of a jarring UI transition when moving in and our of PiP
5. The volume slider thumb is vertically misplaced
6. We should still the UX of the time remaining label from QuickTime. Clicking it changes between total time and time remaining. They also use a `-` to signal that it is the time remaining. Bonus points if we took the tvOS feature of showing what time of day the video will finish :-P


Loving the more SwiftUI, this UI layout is insanely hard to follow.

insidegui:
I'm not discarding the possibility of redoing the whole player in SwiftUI tbh 😅

allenhumphreys:
Surely the controls view could pretty easily be a swiftui view, it's basically just buttons. Maybe a binding or 2. It'd be like 500 fewer lines of code 😂

insidegui:
Something else we can remove (regardless of a SwiftUI migration) is the external playback stuff, since that's been disabled with a compile-time flag for a while.

sugarmo:
Can we add subtitles to downloaded videos? The current situation is that even if a video has subtitles when played online, the subtitles will be lost as long as it is downloaded and played again. In fact, transcript can be used as subtitles. Is it difficult to implement this feature?

insidegui:
@sugarmo That's tracked in #657, feel free to continue this discussion there

insidegui:
> The window width is now constrained to be unnecessarily wide

Fixed in https://github.com/insidegui/WWDC/pull/695/commits/c9379dbe01db8fc8dde9c8843f5928182fc4cbdd

> For the PiP external status view, the status view takes the full space of the view which results in a bit of a jarring UI transition when moving in and our of PiP

Fixed as part of https://github.com/insidegui/WWDC/pull/695/commits/3780e5a11f2b99913b723e3705f69d19b9c0b98f
Specifically: https://github.com/insidegui/WWDC/blob/3780e5a11f2b99913b723e3705f69d19b9c0b98f/PlayerUI/Views/PUIPlayerView.swift#L665-L668

> The volume slider thumb is vertically misplaced

Probably for some weird legacy reason it was using custom drawing, just reverted to a regular `NSSlider` with `.small` control size: https://github.com/insidegui/WWDC/pull/695/commits/8c9bb0cfbc52d7e0bf3949c994de782a4a9a6a62

> We should still the UX of the time remaining label from QuickTime. Clicking it changes between total time and time remaining. They also use a - to signal that it is the time remaining. Bonus points if we took the tvOS feature of showing what time of day the video will finish :-P

Implemented in 0c0fb0aff921d711b7d2b4e282af8baba647f308

I like the idea of having the time of day, will keep that in mind for later :)

> The volume icon could change as the slider changes, e.g. speaker.wave.3.fill -> speaker.wave.2.fill -> speaker.wave.1.fill

Done in 7f3d5a5b4a17f71217e0f10cbaf00ce2124f7d40

> The goforward/backward buttons look squished.

Not sure what happened (maybe different SDKs / macOS versions) but they look fine now, macOS 14.5 building with Xcode 15.3

allenhumphreys:
> We should still the UX of the time remaining label from QuickTime.

Steal* 🤪

Nothing like correcting the internet record 1 year later!

insidegui:
> The book marks indicators a bit funny, could be existing issues

Redesigned in 5d304f46c34c5bef080875a3efd922737f74e7b1

CleanShot 2024-05-29 at 18 24 29

> I managed to create a bunch of book marks all at the same spot, then while removing them I managed to get the app to crash with a constraint not being finite! You can also crash it by just changing the playback speed a few times.

Not sure what the crash was about, but since we don't do clustering or some other solution for now, I've introduced a minimum distance of 30s between bookmarks, which should not be a problem for regular use (I usually have just a few bookmarks in any given session, and they're never really close together like that). Also done in 5d304f46c34c5bef080875a3efd922737f74e7b1

## Extra stuff

- The player view now handles hover tracking for the timeline, reducing the precision required to interact with the timeline for seeking / previewing; this is achieved by expanding the vertical area of the timeline and requiring the cursor to move out further away from the visible area before disengaging the hover interaction
- Restored previous behavior where player controls can be interacted with while the app is in the background (except for the timeline)
- New playback speed control (borrowed the implementation from another branch)
- Cute animation for the play/pause button
- I can't believe this wasn't already the case, but most player settings such as volume, the mode for the time left / duration label, and playback speed are now persisted (a4c4fb8f70c4c3cc06f27092f919c97340afd465, 28f773325c7befe5b3b0de18297511ebaea5d0bb)
- When right-clicking the playback speed control, there's a "Custom" option that allows the user to enter any custom speed between 0.25 and 3.5 (0fd526075a2e345cd68bc9044fd88d5b7c3ea4c2)
- Improved the look of the floating time preview that shows up when hovering over the timeline or editing bookmarks

https://github.com/insidegui/WWDC/assets/67184/989fa16c-04fb-4afa-ae0b-fec6542cd0a0

insidegui:
> I managed to create a bunch of book marks all at the same spot

Mitigated in 9bd28342da58a13df05d93a3ec045b41cf1efe9f

PR #696, opened 2023-06-08
Optimize schedule grouping to avoid excessive CPU and memory usage on start up

allenhumphreys:
After examining the algorithm for creating the schedule groupings, I realized that we only need them sorted by time. The SessionInstance.standardSort is very complex and reaches across linking objects. As a result, large amounts of memory must be used as every object in the graph has to be materialized.

Before:

Screenshot 2023-06-08 at 3 34 37 PM

After:
Screenshot 2023-06-08 at 3 30 43 PM


insidegui:
Amazing work! 🚀

PR #699, opened 2023-06-10
Fixed missing explicit self in iCloud code

insidegui:
There were a couple of missing explicit `self` references after the change to `Logger` in iCloud-only code.

Not sure what `contents.xcworkspacedata` is about 🤔

### DIFF
--- a/WWDC/LiveObserver.swift
+++ b/WWDC/LiveObserver.swift
@@ -221,11 +221,13 @@ private final class CloudKitLiveObserver: Logging {
                                                    subscriptionID: specialLiveEventsSubscriptionID,
                                                    options: options)
 
-            database.save(subscription) { _, error in
+            database.save(subscription) { [weak self] _, error in
+                guard let self = self else { return }
+
                 if let error = error {
-                    log.error("Error creating subscriptions: \(String(describing: error), privacy: .public)")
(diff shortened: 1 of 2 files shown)


allenhumphreys:
I changed the gitignore to pick up the entire "hidden" xcworkspace to allow the resolved file to be committed. That's the one I inadvertently pushed to master. I'm not sure why you have that file and I don't. It's possible it's an artifact of an older Xcode? I did a new clone and only opened the project in Xcode 14 on my machine, so if your clone has been around longer.... maybe?

WWDC/LiveObserver.swift
```diff
@@ -221,11 +221,13 @@ private final class CloudKitLiveObserver: Logging {
                                                    subscriptionID: specialLiveEventsSubscriptionID,
                                                    options: options)
 
-            database.save(subscription) { _, error in
+            database.save(subscription) { [weak self] _, error in
+                guard let self = self else { return }
+
                 if let error = error {
-                    log.error("Error creating subscriptions: \(String(describing: error), privacy: .public)")
+                    self.log.error("Error creating subscriptions: \(String(describing: error), privacy: .public)")
```
allenhumphreys, line 228: Must've been copy paste, the regex converter I built after I completed the migration by hand wouldn't have made that mistake 🤣

allenhumphreys (approved):

PR #700, opened 2023-06-10
Adopt Calendar write-only permission in macOS Sonoma

insidegui:
Since all we do is write to the calendar, we don't need full access.

There's a compile-time check so that the app keeps building just fine in Xcode 14. I wanted something like `#if __MAC_OS_X_VERSION_MAX_ALLOWED` in Swift, but apparently the only way to do a compile-time check for SDK is by using the `#compiler` directive. If you know of a better solution let me know.

CleanShot 2023-06-10 at 10 30 09@2x

### DIFF
--- a/WWDC/AppCoordinator+SessionActions.swift
+++ b/WWDC/AppCoordinator+SessionActions.swift
@@ -69,79 +69,6 @@ extension AppCoordinator: SessionActionsViewControllerDelegate {
         }
     }
 
-    @objc func sessionActionsDidSelectCalendar(_ sender: NSView?) {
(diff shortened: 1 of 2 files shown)


WWDC/AppCoordinator+SessionActions.swift
```diff
@@ -210,3 +137,107 @@ extension Storage {
         setFavorite(!session.isFavorite, onSessionsWithIDs: [session.identifier])
     }
 }
+
+// MARK: - Calendar Integration
```
allenhumphreys, line 141: Separate file? We have lots of AppCoordinator+'s already :)

allenhumphreys (approved):

PR #703, opened 2023-06-22
Sessions table view model refactoring

allenhumphreys:
Simplify how the table data source gets it's data and receives updated results from storage and how that is then combined with filtering predicates.

A structural change that will help make it simpler to convert to diffable datasource and/or leverage SwiftUI or move to outline view, etc. It's just simpler.

- [x] Search coordinator should update filters granularly if a storage change causes options to change and it shouldn't re-apply the saved state
- [x] Search coordinator needs a way to apply the filter selection directly to the filters to support explore deep linking
- [x] App coordinator needs cleaned up and checked for correctness
- [x] WWDC "today" feature needs fixed
- [x] Clearing filters to select a session needs fixed
- [x] Realm schema migration block
- [x] The currently playing video needs to be inserted into filter results like it used to be


Packages/ConfCore/ConfCore/SessionInstance.swift
```diff
@@ -113,17 +112,12 @@ public class SessionInstance: Object, ConditionallyDecodable {
...
-                return newKeyword
-            }
-        }
+        let currentKeywordIds = Set(keywords.map(\.name))
+        other.keywords.forEach { newKeyword in
+            guard !currentKeywordIds.contains(newKeyword.name) else { return }
 
-        keywords.removeAll()
-        keywords.append(objectsIn: otherKeywords)
+            keywords.append(realm.object(ofType: Keyword.self, forPrimaryKey: newKeyword.name) ?? newKeyword)
```
allenhumphreys, line 119: This is no longer equivalent since it doesn't handle removals

Packages/ConfCore/ConfCore/Session.swift
```diff
@@ -323,6 +304,21 @@ extension Session {
         return asset
     }
 
+    public func thumbImageAsset() -> SessionAsset? {
```
allenhumphreys, line 307: Remove this

Packages/ConfCore/ConfCore/Session.swift
```diff
@@ -129,36 +117,29 @@ public class Session: Object, Decodable {
...
-            }
-
-            guard !related.contains(where: { $0.identifier == effectiveRelated.identifier }) else { return }
-            related.append(effectiveRelated)
+            related.append(realm.object(ofType: RelatedResource.self, forPrimaryKey: newRelated.identifier) ?? newRelated)
         }
 
+        let currentFocusIds = Set(focuses.map { $0.name })
         other.focuses.forEach { newFocus in
-            let effectiveFocus: Focus
```
allenhumphreys, line 151: The existing logic doesn't handle removals, should we fix that?

allenhumphreys: (that's true for all 3 lists that get updated here)

Packages/ConfCore/ConfCore/Storage.swift
```diff
@@ -542,26 +614,57 @@ public final class Storage: Logging {
...
+            }
     }
 
     public var allTracks: [Track] {
         return realm.objects(Track.self).sorted(byKeyPath: "order").toArray()
     }
 
 }
+
+public extension RealmCollection where Self: RealmSubscribable {
```
allenhumphreys, line 640: Move to a new file

Packages/ConfCore/ConfCore/Storage.swift
```diff
@@ -542,26 +614,57 @@ public final class Storage: Logging {
...
+                case .update:
+                    return nil
+                case .error(let error):
+                    throw error
+                }
+            }
+    }
+}
+
+private func merge<T>(old: List<T>, new: List<T>) {
```
allenhumphreys, line 660: Remove

Packages/ConfCore/Package.swift
```diff
@@ -14,6 +14,7 @@ let package = Package(
             targets: ["ConfCore"])
     ],
     dependencies: [
+        .package(url: "https://github.com/apple/swift-collections.git", from: "1.0.4"),
```
allenhumphreys, line 17: Not in use anymore

insidegui (approved):
Looking good! Haven't found any issues living on this build for a couple of days :)

PR #715, opened 2024-06-12
Complex filters

### LINKED ISSUE #583: Ability to filter for Non-Downloaded videos.
Hopefully this desired functionality can adequately be described as being the opposite of what the 'Downloaded' filter provides. Essentially displaying what videos have not yet been been download, allowing one to hone in on what videos may want to be added to the download queue.

This would be in addition to the existing downloaded indicator within each list cell, and would most likely (hopefully) have no bearing on this nicety.

davedelong:
This adds complex filters to the video list, so users can choose to (among other things) filter to non-downloaded videos.

Fixes #583


davedelong:
Screenshot 2024-06-11 at 7 22 47 PM

allenhumphreys:
I built and ran this. Looks pretty straightforwardly implemented.

I wish we had a better control for this, but until we get the filters out of the storyboard, doing anything great especially using SwiftUI is a real uphill battle.

Don't let perfect be the enemy of a good feature, though. We could bikeshed the words in the segmented controls if we wanted to, but I think we're all pretty busy :-)

allenhumphreys (approved):

insidegui:
Thanks @davedelong 🥳

PR #717, opened 2024-06-20
Add playback speed change shortcuts with < and > keys.

hvsw:
Like YouTube does it. Same idea as https://github.com/insidegui/WWDC/pull/581

Screenshot 2024-06-20 at 12 07 47


insidegui:
Thanks @hvsw <3

PR #726, opened 2025-07-11
Convert RelatedSesssionsViewController to SwiftUI

allenhumphreys:
- Converts related sessions, building on the previous PR to convert the cells
- Updates minimum deployment target to 14.2 (avoiding low number point releases while still keeping on 14, for now)
- Fixes a performance issue while resizing the window or Split View
- Gets rid of some new deprecation warnings
- Deleted unused file that contained warnings

image

Refactoring was planned an partially executed by Claude Code. But I've thoroughly tested it!

### DIFF
--- a/WWDC/RelatedSessionsViewController.swift
+++ b/WWDC/RelatedSessionsViewController.swift
@@ -55,13 +55,14 @@ struct RelatedSessionsView: View {
                     HStack(spacing: Metrics.itemSpacing) {
                         ForEach(viewModel.sessions, id: \.identifier) { session in
                             sessionButton(for: session, in: scrollView)
+                                .id(session.identifier)
                         }
                     }
                     .padding(.horizontal, 0)
                     .padding(.bottom, Metrics.scrollerOffset)
                 }
-                .onChange(of: viewModel.sessions.map(\.identifier)) { _, _ in
-                    scrollView.scrollTo(viewModel.sessions.first?.identifier, anchor: .leading)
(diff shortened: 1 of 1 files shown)


WWDC/CursorShapeViewModifier.swift
```diff
@@ -0,0 +1,97 @@
...
+
+import AppKit
+import SwiftUI
+
+/// Inside of a scroll view, the push/pop behaves not so great when re-capturing the hover after the scroll view
+/// scrolls. Scrolling the scroll view doesn't cancel the hover but it does change the cursor back to an arrow.
+///
+/// It may be possible to work around this via onContinuousHover and some cancelling delay shenanigans,
+/// but it seems like a lot of effort for a minor issue.
+struct CursorShapeViewModifier: ViewModifier {
```
insidegui, line 17: This is a neat modifier! I couldn't find where it's being used though. Am I missing something?

allenhumphreys: @insidegui  I created this ahead of time to replace ActionLabel which is actually mostly used by the About window. ActionLabel is just text you can click and the pointing hand makes that clear. But there's one ActionLabel in the session summary... but I think it's broken so you maybe forgotten it's there 😂

allenhumphreys:
That's my bad. I had it working, then made it "better" and failed to re-test it 🤦🏻‍♂️


https://github.com/user-attachments/assets/3a88fc2e-d850-4166-b99d-1f7a04c6c8b5

PR #728, opened 2025-07-17
Convert session actions to SwiftUI

allenhumphreys:
Converts the session actions view controller (we really needed an entire VC for some buttons... :joy:) to SwiftUI.

FYI, I've noticed some odd behaviors with the download manager that I don't think are related to any changes in this PR. But I haven't gone looking to understand them. Mostly around pausing, cancelling and completing downloads. Also downloading, then deleting, then re-downloading.

This PR changes the rhs of the app so that it is split evenly vertically, reducing UI jumping around when changing sessions. It's not completely perfect, but none of this really is. The goal is to get to a point where changing the UI is no longer a herculean task.

https://github.com/user-attachments/assets/d0968e76-dd8c-4816-964d-7c8c6f3985f6

Code changes here were helped along by Claude Code and GitHub Copilot for Xcode, but I did the important work :-)


insidegui:
> I've noticed some odd behaviors with the download manager that I don't think are related to any changes in this PR. But I haven't gone looking to understand them. Mostly around pausing, cancelling and completing downloads. Also downloading, then deleting, then re-downloading.

If you can, please document those in an issue.

I did a lot of work related to downloads in #713, so it's possible that these issues were caused by that. I haven't noticed any issues with downloads during regular use of the app, but that could be because I simply never pause/resume downloads. Speaking of which, why do we need to support manually pausing/resuming downloads? 🤔

PR #729, opened 2025-07-18
Converted SessionSummaryViewController to SwiftUI

allenhumphreys:


PlayerUI/Views/PUIButtonView.swift
```diff
@@ -57,18 +57,18 @@ import SwiftUI
 /// **Future Direction**: This will eventually be replaced by a pure SwiftUI implementation
 /// once all complex AppKit dependencies are resolved.
 public struct PUIButtonView: View {
-    @State private var button: PUIButton
+    @StateObject private var button: PUIButton
```
allenhumphreys, line 60: This is what I meant to do to begin with. The before state was leaking PUIButton's.

WWDC/ImageDownloadCenter.swift
```diff
@@ -34,7 +34,29 @@ final class ImageDownloadCenter: Logging {
     
     func cachedThumbnail(from url: URL) -> NSImage? { cache.cachedImage(for: url, thumbnailOnly: true).thumbnail }
 
-    /// The completion handler is always called on the main thread
+    func downloadImage(
```
allenhumphreys, line 37: This is sorta hacky meant to be a POC, but it mostly gets the job done.

WWDC/ImageDownloadCenter.swift
```diff
@@ -199,7 +221,21 @@ final class ImageCacheProvider {
             key = sourceURL.imageCacheKey as NSString
         }
 
-        guard fileManager.fileExists(atPath: url.path), let image = NSImage(contentsOf: url) else { return nil }
+        guard
+            fileManager.fileExists(atPath: url.path),
+            let data = try? Data(contentsOf: url),
+            let source = CGImageSourceCreateWithData(data as CFData, nil),
```
allenhumphreys, line 227: This is an attempt to pre-decode the image, I don’t think it is actually working.

WWDC/RelatedSessionsViewController.swift
```diff
@@ -54,36 +54,49 @@ struct RelatedSessionsView: View {
...
     }
 
-    func sessionButton(for session: SessionViewModel, in scrollView: ScrollViewProxy) -> some View {
+    func sessionButton(for session: SessionCellViewModel, in scrollView: ScrollViewProxy) -> some View {
         Button {
-            viewModel.selectSession(session)
+            viewModel.selectSession(session.viewModel!)
         } label: {
             SessionCellView(
-                cellViewModel: SessionCellViewModel(session: session),
```
allenhumphreys, line 86: The outer VM needed to own this, SessionCellView is just @ObservedObject. Partly because of the table view.

WWDC/Realm+Combine.swift
```diff
@@ -11,18 +11,30 @@ import RealmSwift
...
 
+        // by using `.multicast(subject: CurrentValueSubject(self))` we are sharing the upstream publisher
+        // and also replaying the most recent value to new subscribers which is useful when
+        // subscriptions come in at slightly different times
         switch (share, includeInitialValue) {
         case (true, true):
-            return Publishers.Concatenate(prefix: initialValue, suffix: valuePublisher.share()).eraseToAnyPublisher()
+            return valuePublisher
+                .prepend(self)
+                .multicast(subject: CurrentValueSubject(self))
```
allenhumphreys, line 23: I don’t think I need both prepend and the current value subject.

CurrentValueSubject always emits its most recent value when subscribing to it. This has the effect of being a multicast (one realm subscription) that will replay its latest value to subscribers. 

But I really need to double check that. Ultimately this is just a more nuanced version of the share() operator.

PR #732, opened 2025-07-28
Use realm-swift with Realm.xcframework binaryTarget

allenhumphreys:
# Motivation

1. realm-core takes a long time to build
2. SPM doesn't let us specify that for a particular dependency we do NOT want a debug build
3. Realm is VERY VERY slow when built for debug, which is fine for them, but when developing the app, it leads to ambiguity as to whether our implementation of realm is slow, or if it's just the realm debug build.

# Solution

Use Realm & realm-core binary dependencies!

## Explanation

realm-swift has 2 products:
1. Realm (realm-core is statically linked)
4. RealmSwift

Realm depends on realm-core and is written in Objective-C++. 

RealmSwift depends on Realm.

After the versions are released, Realm is available as a binary on the realm-swift release. Swapping in that binary means we will get release build of Realm and realm-core. RealmSwift will still be built from source because they're not using library evolution or whatever which means you need a specific swift compiler binary output to match to your Xcode version. RealmSwift is actually a relatively thin wrapper around the other 2 libraries and doesn't contribute a lot to the performance ambiguity.

Other approach explored in #730

### DIFF
--- a/Packages/ConfCore/Package.swift
+++ b/Packages/ConfCore/Package.swift
@@ -16,7 +16,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/apple/swift-collections.git", from: "1.0.4"),
         .package(url: "https://github.com/bustoutsolutions/siesta", from: "1.5.2"),
-        .package(url: "https://github.com/realm/realm-swift", from: "10.0.0"),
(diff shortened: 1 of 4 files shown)


insidegui:
It's amazing how much faster clean builds are with this patch ❤️

PR #735, opened 2025-07-31
[TODO] Fix the incorrect detached status display

bo2themax:
While adopting Liquid Glass, I’ve noticed an issue with the detached status. I tested it with the release version, and it appears to be a historical problem.

**When a video enters PiP mode and a new session is selected, the new session thumbnails are displayed as the status**. This should be the corresponding thumbnails as the video is not currently playing. Switching back to the playing video displays the default status view provided by AVPlayer which should be the custom status view attached to the incorrect session.

https://github.com/user-attachments/assets/15ee2398-a38f-453f-9b6f-58d43ab5662f

---

Debugging revealed multiple `$selectedSession` listeners. One updates the view model of the detail view and the other checks if it is appropriate to enter PiP mode. However, the order of these two listeners is not deterministic so they can occur in any order. If the PiP listener occurs early it should function as expected. However, as far as I have tested on macOS Tahoe, after opening WWDC for the first time they almost appeared in the incorrect order in 90% of cases.

This fix comprises three stages, each addressing the previous one:

1. Modify the update order to ensure the second listener is executed.
2. Resolve the issue where a new video replaces the Picture-in-Picture content without altering the detached status of that session.
3. Correct the positioning of the detached status.
4. Add a status check we attaching detected status indicator

The result behaviour is looking like this now. I’ve tried to stick as closely to the codebase as possible, but there might still be a few spots where a change could make things even better.🙈

https://github.com/user-attachments/assets/52cd1602-8cf0-45a0-aff2-3e27fc8206c6

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -20,6 +20,10 @@ public final class PUIPlayerView: NSView {
 
     private let log = Logger(subsystem: "PlayerUI", category: "PUIPlayerView")
     private var cancellables: Set<AnyCancellable> = []
+    /// Bindings that only need to be established once and are not dependent on the player (model).
+    /// The nature of Combine forces us to store them somewhere, though. And as long as they only capture `self`
+    /// weakly, the ui bindings don't need to be reset when the player changes.
+    private var uiBindings: Set<AnyCancellable> = []
 
     // MARK: - Public API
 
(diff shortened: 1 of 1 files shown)


bo2themax:
> I noticed this problem, too!

Found some issues for this fix when playing in fullscreen mode, made it WIP for now

allenhumphreys:
Be careful, I have a relatively large refactor going on in #729 . I'm hoping to get back to it soon, but I don't want to end up with lots of conflicts!

bo2themax:
> Be careful, I have a relatively large refactor going on in #729 . I'm hoping to get back to it soon, but I don't want to end up with lots of conflicts!

Added a check when attaching the indicator view, please review the changes if they have any potential conflicts with your refactor

PR #747, opened 2025-09-05
Fix PiP return animation via swizzling!

allenhumphreys:
- Fix PiP return animation via swizzling the method that will return the correct frame
- Fix enter full screen animation by detecting window visibility before entering full screen

### DIFF
--- a/PlayerUI/Views/PUIPlayerView.swift
+++ b/PlayerUI/Views/PUIPlayerView.swift
@@ -13,6 +13,7 @@ import AVKit
 import Combine
 import SwiftUI
 import ConfUIFoundation
+import ObjectiveC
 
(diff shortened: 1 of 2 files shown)


PlayerUI/Views/PUIPlayerView.swift
```diff
@@ -1652,6 +1672,63 @@ extension PUIPlayerView: AVPictureInPictureControllerDelegate {
...
+        lock.withLock {
+            _pipReturnDestination = rect
+
+            whileLocked()
+
+            _pipReturnDestination = nil
+        }
+    }
+
+    @available(macOS, deprecated: 26, message: "Check to make sure this still works, then raise the deprecation")
```
insidegui, line 1694: 👌🏻

PR #748, opened 2025-09-07
Get IDEProvisioningTeamByIdentifier first when scanning existing teams

bo2themax:
IDEProvisioningTeams appears to be deprecated for a clean install of Xcode 26.


allenhumphreys (approved):
The new method works great, but we should probably document that the top method is the "new way" and that the bottom section can be removed after support for Xcode < 26 is dropped from the project.

allenhumphreys (approved):

PR #750, opened 2025-09-11
Update Project Settings for Contributors

bo2themax:
[647ca95](https://github.com/insidegui/WWDC/commit/647ca950a2a50dc65a5d79a1809f7fbca71771d1) accidentally committed a different TeamID, resulting in Xcode not reading the team id from the Xcode configuration, so:

- Remove DEVELOPMENT_TEAM from xcodeproj
  > Removing this can prevent Xcode from modifying "$(inherited)" to the derived team ID when users select the signing tab, to avoid unintended changes.

- Update Package.resolved for ConfCore, by the way, for [8f31557](https://github.com/insidegui/WWDC/commit/8f31557d3ba8e7cb1038eb5e8de3a3519471ad57)


WWDC.xcodeproj/project.pbxproj
```diff
@@ -822,7 +822,6 @@
 				COPY_PHASE_STRIP = NO;
 				DEAD_CODE_STRIPPING = YES;
 				DEBUG_INFORMATION_FORMAT = "dwarf-with-dsym";
-				DEVELOPMENT_TEAM = 72X6ZPCY5L;
```
allenhumphreys, line 825: I had previously set these to $(inherited) because Xcode kept trying to copy the value from the config into this file whenever you change a build setting through the UI. At least I’m pretty sure that got merged in. 

Did Xcode overwrite $(inherited) or was there a conflict resolution where you took your side?

allenhumphreys: Hmmm, it looks like this was actually my fault! Gonna check on that more today.

bo2themax: > I had previously set these to $(inherited) because Xcode kept trying to copy the value from the config into this file whenever you change a build setting through the UI. At least I’m pretty sure that got merged in. 
> 
> Did Xcode overwrite $(inherited) or was there a conflict resolution where you took your side?

Yeah, I tried using $(inherited), but once I opened the signing tab, and switched to the repo, actual id was overwritten

allenhumphreys: 😞 That's very frustrating! I’ve had this problems with other settings in xcconfig files before, too. I should probably take that time to create a sample project and open a Feedback.

bo2themax: > 😞 That's very frustrating! I’ve had this problems with other settings in xcconfig files before, too. I should probably take that time to create a sample project and open a Feedback. 

I think it's been like this for a long time, almost every project I cloned has last team id embedded😅

bo2themax:
@allenhumphreys  This pr indeed does not fix the issue. 

It seems to me that a pattern exists here. As long as you don’t open WWDC under PROJECT settings, no new changes will be added.

But another project using Xcode configuration to distinguish bundle id doesn't have this behaviour, which is very strange.🥲

https://github.com/user-attachments/assets/486a4b0e-4456-44b6-9b9a-90d0fa322242

allenhumphreys:
> @allenhumphreys This pr indeed does not fix the issue.
> 
> It seems to me that a pattern exists here. As long as you don’t open WWDC under PROJECT settings, no new changes will be added.
> 
> But similar projects using Xcode configuration to distinguish bundle id doesn't have this behaviour, which is very strange.🥲

Yes, this particular setting (`DEVELOPMENT_TEAM`) seems to be treated differently.

A potential solution/workaround/defensive measure I thought of was to add a build phase script that removes these lines using sed. I don't know if writing the pbxproj file during a build will cause a major problems though.

allenhumphreys:
Also wow, good job figuring out exactly the scenario in which it does it. It doesn't do it if you open target settings. Or even signing settings in the target, but as soon as you open the project level settings it overwrites it!

allenhumphreys:
A working sed script/command `sed -i '' '/^[[:blank:]]*DEVELOPMENT_TEAM[[:blank:]]*=[[:blank:]]*[A-Za-z0-9]*;[[:blank:]]*$/d' WWDC.xcodeproj/project.pbxproj`.

Xcode is 100% doing the wrong thing though.

allenhumphreys:
I tested this out with 26 and it didn't seem to mind too much. Emitting the error and failing the build is somewhat optional, but I want to be informed!

image

```bash
if ! sed '/^[[:blank:]]*DEVELOPMENT_TEAM[[:blank:]]*=[[:blank:]]*[A-Za-z0-9]*;[[:blank:]]*$/d' WWDC.xcodeproj/project.pbxproj | cmp -s - WWDC.xcodeproj/project.pbxproj; then
    echo "error: Had to remove DEVELOPER_TEAM from the pbxproj file, build again"
    sed -i '' '/^[[:blank:]]*DEVELOPMENT_TEAM[[:blank:]]*=[[:blank:]]*[A-Za-z0-9]*;[[:blank:]]*$/d' WWDC.xcodeproj/project.pbxproj
    return 1
fi
```

bo2themax:
@allenhumphreys 

- Moved all the scripts in the scripts folder and updated README
- Added the script to pre-commit hooks
- Added a step in bootstrap to set this up

Adding the script in the build phase, I think, can't avoid developers committing this change, but git hooks will warn them.


https://github.com/user-attachments/assets/c1907b9b-7b3a-46cb-ba20-851fe7df3bb0

